PR #16, opened 2021-08-24
Fixing LSP1, LSP3, LSP4, LSP6

YamenMerhi:
Changes in LSP1 : 
- Added Interface Cheat Sheet

Changes in LSP3 :
- Fixed outdated links

Changes in LSP4 : 
- Fixed outdated links
- Added Interface Cheat Sheet

Changes in LSP6 : 
- Edited outdated Interface Cheat Sheet
- Fixed outdated links


YamenMerhi:
Some events are emitted with private functions so it will be less attractive if we put events related to public functions underneath and leave the events that are related to private function at the top, that's why i made all the events at the top.
It's up to you @frozeman

PR #21, opened 2021-09-22
Update LSP-1-UniversalReceiver.md

luki3k5:
I added some small changes (incl. embed of the image that explains this in within a single picture from yesterday's presentation). After all an image is worth a thousand words ;)


frozeman:
sadly this image is a bit misleading, as ERC20 tokens are not directly compatible, rather the new LSP7 and LSP8 that will come. LSP7 is a ERC20-like token standard, that adds the LSP1.

the image would need to be adjusted and made also smaller, so it doesnt take so much space. but i am personally not fully convinced if its the right pic :)

frozeman:
closing for now

PR #22, opened 2021-09-23
Order of params for executeRelayCall + Fixed Interface Sheetcheat

CJ42:
- [x] changed order of `executeRelayCall` function parameters. `_data` moved from 1st to 3rd parameter.
- [x] fixed error in LSP2 for `**AddressMappingWithGrouping** (incorrect value for `bytes2(SecondWord)`)
- [x] Added `_channel` parameter for `getNonce` in **Interface Sheetcheat** 

@YamenMerhi @frozeman , for the function `getNonce(...)` (in both the **Methods** and **Interface SheetCheat**), should we put the`_channel` parameter as `uint128` instead of `uint256`?
This would align with what is mentioned above the figure in [**Nonces in the KeyManager**](https://github.com/lukso-network/LIPs/blob/master/LSPs/LSP-6-KeyManager.md#nonces-in-the-keymanager)

### DIFF
--- a/LSPs/LSP-6-KeyManager.md
+++ b/LSPs/LSP-6-KeyManager.md
@@ -248,7 +248,8 @@ The Key Manager allows out-of-order execution of messages by using nonces throug
  - left most 128 bits : `channelId`
  - right most 128 bits: `nonceId`
 
-![multi-channel-nonce](https://user-images.githubusercontent.com/86341666/31145285/133279354-82bebc4f-21f4-40e4-b959-93ccd624e5c4.jpg)
+
+![multi-channel-nonce](https://user-images.githubusercontent.com/31145285/133292580-42817340-104e-48c5-832b-533842b98d26.jpg)
 
 <p align="center"><i> Example of multi channel nonce, where channelId = 5 and nonceId = 1 </i></p>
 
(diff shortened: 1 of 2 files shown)


frozeman:
i agree channel should be uint128

LSPs/LSP-6-KeyManager.md
```diff
@@ -335,11 +335,11 @@ interface ILSP6  /* is ERC165 */ {
     event Executed(uint256 indexed  _value, bytes _data); 
     
     
-    function getNonce(address _address) external view returns (uint256);
+    function getNonce(address _address, uint256 _channel) external view returns (uint256);
     
     function execute(bytes calldata _data) external payable returns (bool);
     
-    function executeRelayCall(bytes calldata _data, address _signedFor, uint256 _nonce, bytes memory _signature) external payable returns (bool);
+    function executeRelayCall(address _signedFor, uint256 _nonce, bytes calldata _data, bytes memory _signature) external payable returns (bool);
```
frozeman, line 342: maybe e memory is better here too as this function could also be executed by another smart contract forwarding a signature

frozeman (approved):

PR #27, opened 2021-09-29
Token standards

mattgstevens:
### Description
Create the standards for LSP7 and LSP8, and updates to LSP4 based on conversations with lukso and fanzone teams.

LSP7 and LSP8 will be the token standards, and LSP4 defines the common metadata.


mattgstevens:
@frozeman changed the names for LSP7 and LSP8 to use "DigitalAsset" so that LSP4 remains as "DigitalCertificate".

possibly you have some opinion here on the naming. if we want to keep these all as "DigitalCertificate" maybe we change LSP4 to "DigitalCertificateMetadata" or something similar.

PR #29, opened 2021-10-12
docs: fix Bytes20Mapping

YamenMerhi:
- Fix Bytes20Mapping
- Fix calldata to memory
- Fix title from Key to ERC725Key
- to Italic text


YamenMerhi:
@CJ42 Once the KeyManager is finished we can set the return to bytes not bool

PR #34, opened 2021-11-09
LSP6 - write permissions as bytes32 with zero left-padding

CJ42:

- [x] Correct the code snippet for each permissions on LSP6.
- So that they can be used via copy-pasting ✅ 

Permissions values for `AddressPermissions:Permissions:<address>` must be written as `bytes32` long + **zero left-padded**

The current examples show them as `0x0004`.

If these values are used via copy-pasting in code snippets, they will be zero right-padded, and will lead to odd or difficult to debug revert errors from the Key Manager.

### DIFF
--- a/LSPs/LSP-6-KeyManager.md
+++ b/LSPs/LSP-6-KeyManager.md
@@ -1,4 +1,4 @@
----
+****---
 lip: 6
 title: Key Manager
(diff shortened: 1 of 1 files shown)


frozeman (approved):

frozeman:
PS. please sign your commits

PR #44, opened 2021-11-30
Use index instead of number for Array

Hugoo:


frozeman:
thats not correct, as its the number of elements not the index of one single item

Hugoo:
ah indeed no big deal. I think the first occurrence is `number` but the other 2 are `index`.

PR #54, opened 2022-01-05
refactor: use `bytes4` for `SupportedStandards` key

CJ42:

The ERC725Y keys `SupportedStandards:{StandardName}` are defined with `bytes` as their `valueType`

However, these keys all use a `bytes4` literal as a value `valueContent`

Change the LSP Metadata specs that use these key by specifying `bytes4`  as `valueType`


CJ42:
@frozeman awaiting to rebase after PR #53 is merged.

Hugoo:
We should also update the JSON files here for erc725.js:

https://github.com/ERC725Alliance/erc725.js/blob/develop/src/schemas/LSP4DigitalAsset.json

Hugoo (approved):

frozeman (approved):

PR #71, opened 2022-04-05
feat(lsp6)!: add `account()` function in LSP6 specs

CJ42:

## Features

- add the function `account()` as part of the standard interface of LSP6

## Docs

- small change in LSP2 Summary
- add missing ERC165 interface IDs in other LSPs


JeneaVranceanu (changes requested):

LSPs/LSP-6-KeyManager.md
```diff
@@ -182,6 +186,16 @@ SIGN               = 0x000000000000000000000000000000000000000000000000000000000
 
 ### Methods
 
+#### account
+
+```solidity
+function account() external view returns (address)
```
frozeman, line 192: I think we should rename this to `target()`

PR #72, opened 2022-04-15
fix: errors in JSON schemas of keys

CJ42:

All of the ERC725Y JSON Schema of the keys with a `Map` keyword at the end (or in other words that have `<address>`:

- are defined as `Mapping` ❌ 
- but should `Bytes20Mapping` ✅


YamenMerhi (approved):
Great!

frozeman (approved):

PR #89, opened 2022-06-08
Use tuple of instead of `Mixed` for LSP2 Map data keys

CJ42:

## Current issue

Currently most Map data keys schemas define the valueContent as `Mixed`.

This does not provide context on how to decode and interpret the raw underlying bytes stored under such keys, as these could be of any types.

## Suggested changes

1. add description about the possibility of using tuple of types for:

- `valueType` = _e.g._ `(bytes4,bytes8)`
- `valueContent` = _e.g._ `(Bytes4,Number)`

2. change Map keys in LSP4, LSP5, LSP10 and LSP12 to use tuples for `valueType` and `valueContent`


YamenMerhi (approved):

YamenMerhi:
LSP8 Schemas to be discussed later

PR #91, opened 2022-06-10
Add `SupportedStandards:LSP9Vault` data key

YamenMerhi:
- Add `SupportedStandards:LSP9Vault` data key


frozeman:
there are no specific LSP9 data keys, so not sure if we need that (for now)

PR #97, opened 2022-07-11
Add missing Boolean type on LSP-2

Hugoo:
Line [62](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md?plain=1#L62) says there is `Boolean` but `Boolean` was missing from the main content table.
Either we delete from L62 or we add it as per this PR.


CJ42 (approved):
great catch!

PR #102, opened 2022-08-02
chore: updated permissions declaration/description format

JeneaVranceanu:
# What this PR introduces? 

Updated block of JS code where permissions' names and values are declared, with related documentation.
Moved JS docs for permissions to be on top of each permission constant, not on the same line as the constant as these docs were almost invisible.

Screenshot 2022-08-02 at 11 56 52

Other related topics to discuss are:
1) the naming of `SUPER_...` permissions;
    Suggest removing `SUPER_` prefix from permissions that have no limitations and adding prefix to permissions that **represent a subset** of current `SUPER_` permissions. 
2) Image with an example of permissions - looks like Alice doesn't have SUPER permissions and the total value of assigned permissions `0x...ffff` shown on the right side of the image is incorrect? In [`lsp-smart-contracts`](https://github.com/lukso-network/lsp-smart-contracts/blob/develop/constants.js#L170) the default `ALL_PERMISSIONS` value is `0x7FBF` which includes all permissions except `DELEGATECALL`, I suppose for security reasons. Bits for permissions that are not declared in this standard are set to 0 in constants in lsp-smart-contracts repository (link is above) but on the image, it feels like all 32 bytes are set to `0xF`. As I recall all bytes were `0xF` initially but later updated to zero out all permissions that do not exist in the standard yet.

### DIFF
--- a/LSPs/LSP-6-KeyManager.md
+++ b/LSPs/LSP-6-KeyManager.md
@@ -174,7 +174,7 @@ CHANGEOWNER         = 0x00000000000000000000000000000000000000000000000000000000
 // Allows changing the permissions (adding + removing) of addresses
 CHANGEPERMISSIONS   = 0x0000000000000000000000000000000000000000000000000000000000000002;
 // .... .... .... 0100 
-// Allows adding new permissions to addresses (removing permission disallowed) 
+// Allows adding new controller addresses by granting them some permissions
 ADDPERMISSIONS      = 0x0000000000000000000000000000000000000000000000000000000000000004;
 // .... .... .... 1000 
 // Allows setting data on the controlled contract
--- a/LSPs/LSP-6-KeyManager.md
+++ b/LSPs/LSP-6-KeyManager.md
@@ -183,7 +183,7 @@ SETDATA             = 0x00000000000000000000000000000000000000000000000000000000
 // Allows calling other contracts through the controlled contract
 CALL                = 0x0000000000000000000000000000000000000000000000000000000000000010;
 // .... .... 0010 .... 
-// Allows calling other contracts through the controlled contract
+// Allows static calling other contracts through the controlled contract, while restrcting any state modifications during the call (or any subcalls, if present)
 STATICCALL          = 0x0000000000000000000000000000000000000000000000000000000000000020;
 // .... .... 0100 .... 
 // Allows delegate calling other contracts through the controlled contract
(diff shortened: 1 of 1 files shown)


LSPs/LSP-6-KeyManager.md
```diff
@@ -167,21 +167,51 @@ By setting the value to `0xeafec4d89fa9619884b6000000000000000000000000000000000
...
+// Allows calling other contracts through the controlled contract
+CALL                = 0x0000000000000000000000000000000000000000000000000000000000000010;
+// .... .... 0010 .... 
+// Allows calling other contracts through the controlled contract
+STATICCALL          = 0x0000000000000000000000000000000000000000000000000000000000000020;
+// .... .... 0100 .... 
+// Allows delegate calling other contracts through the controlled contract
+DELEGATECALL        = 0x0000000000000000000000000000000000000000000000000000000000000040;
+// .... .... 1000 .... 
+// Allows deploying other contracts through the controlled contract
```
CJ42, line 192:
```suggestion
// Allows deploying new contracts through the controlled contract
```

LSPs/LSP-6-KeyManager.md
```diff
@@ -167,21 +167,51 @@ By setting the value to `0xeafec4d89fa9619884b6000000000000000000000000000000000
...
+// Allows adding new permissions to addresses (removing permission disallowed) 
+ADDPERMISSIONS      = 0x0000000000000000000000000000000000000000000000000000000000000004;
+// .... .... .... 1000 
+// Allows setting data on the controlled contract
+SETDATA             = 0x0000000000000000000000000000000000000000000000000000000000000008;
+// .... .... 0001 .... 
+// Allows calling other contracts through the controlled contract
+CALL                = 0x0000000000000000000000000000000000000000000000000000000000000010;
+// .... .... 0010 .... 
+// Allows calling other contracts through the controlled contract
```
CJ42, line 186:
```suggestion
// Allows static calling other contracts through the controlled contract, while restrcting any state modifications during the call (or any subcalls, if present)
```

CJ42 (changes requested):
NIce @JeneaVranceanu ! The comments like this looks much better indeed 👌 
I have added some comments.

CJ42:
> image with an example of permissions - looks like Alice doesn't have SUPER permissions and the total value of assigned permissions 0x...ffff shown on the right side of the image is incorrect? In [lsp-smart-contracts](https://github.com/lukso-network/lsp-smart-contracts/blob/develop/constants.js#L170) the default ALL_PERMISSIONS value is 0x7FBF which includes all permissions except DELEGATECALL, I suppose for security reasons.

That's a good point, will update to clarify it.

CJ42:
@frozeman 

> By the way to change permission permission should not allow adding permissions. Only update existing ones, or removing them.

So like this? ⬇️ 

Screenshot 2022-08-11 at 08 02 51

CJ42:
@frozeman reviewed this PR. This can be merged ✅

frozeman (approved):

PR #103, opened 2022-08-02
Change `LSP3IssuedAssets` to `LSP12IssuedAssets`

b00ste:
The text in the LSP2 standard is outdated LSP3issuedAssets should be replaced with LSP12, and the keys reference.


YamenMerhi (approved):
Cool

PR #104, opened 2022-08-09
Change NFT to non-divisible

b00ste:
Assuming that LSP7 is for tokens like ERC20, not NFT like LSP8 it might be confusing in sense that some might think that setting isNFT to true would create an NFT, which is not entirely true. In this sense the param should be rather called isNonDivisible.


YamenMerhi (approved):
Cool

PR #107, opened 2022-08-15
Update LSP4 Metadata with examples

jxyang44:
Added common metadata classification examples to help streamline front-end tools


frozeman (approved):

frozeman:
Maybe COULD would be better, but merging this for now.

PR #114, opened 2022-08-24
Typo correction

EngrGord:


frozeman:
This fix was already applied as it seems, as no change was suggested

PR #121, opened 2022-08-31
Use `uint8` instead of `uint256` decimals in LSP7

YamenMerhi:
- Use `uint8` instead of `uint256` in LSP7 decimals

We don't need `uint256` for decimals, `uint8` is more than enough. With `uint8`, decimals can be used up to 255.
Most implementations use 18 as a decimal number.


b00ste (approved):

CJ42 (approved):
nice change 👍

PR #125, opened 2022-09-20
fix: SUPER permission bytes positioning fixed

JeneaVranceanu:
SUPER permissions are set as the left-most bits of permissions value.


JeneaVranceanu:
Permissions ordering changed. These changes are irrelevant.

PR #133, opened 2022-10-05
LSP-15-TransactionRelayerAPI

Hugoo:


LSPs/LSP-15-TransactionRelayerAPI.md
```diff
@@ -0,0 +1,92 @@
...
+created: 2022-10-05
+---
+
+## Simple Summary
+
+A Transaction Relayer API for consistency across all Transaction Relayer providers.
+
+## Abstract
+
+The [LSP-6-KeyManager](./LSP-6-KeyManager.md) proposes an [`executeRelayCall()`](./LSP-6-KeyManager.md#executerelaycall) function. It allows anybody to execute `_calldata` payload on a set ERC725 X or Y smart contract, given they have a signed message from an executor. This opens the way to Transaction Relayers. This document describes the API
```
CJ42, line 17:
```suggestion
The [LSP-6-KeyManager](./LSP-6-KeyManager.md) proposes an [`executeRelayCall()`](./LSP-6-KeyManager.md#executerelaycall) function. It allows anybody to execute `_calldata` payload on a set ERC725 X or Y smart contract, given they have provided a signed message to an executor. This opens the way to Transaction Relay service. 

This standard describes the API fro a transaction relay service.
```

LSPs/LSP-15-TransactionRelayerAPI.md
```diff
@@ -0,0 +1,92 @@
...
+{
+  "transactionHash": "0xBB645D97B0c7D101ca0d73131e521fe89B463BFD",
+}
+```
+
+#### POST `/quota`
+
+Returns the available quota left for a registered Universal Profile.
+
+- `signature` value is the message value signed by a controller key with the SIGN permission of the Universal Profile
```
CJ42, line 57: why the `SIGN` permission would be necessary here?

LSPs/LSP-15-TransactionRelayerAPI.md
```diff
@@ -0,0 +1,92 @@
...
+- `timestamp` must be +/- 5 seconds
+- hash should be calculated as `keccack256(address, timestamp)`. This is the message that should be signed.
```
CJ42, line 59:
```suggestion
- hash should be calculated as `keccak256(address, timestamp)`. This is the message that should be signed.
```

Also it might be good to clarify the type for `timestamp` here (I guess a `uint256`?).

So that would be:

```suggestion
- hash should be calculated as `keccack256(address signer + uint256 timestamp)`. This is the message that should be signed.
```

JeneaVranceanu:
It would probably also be worth changing `...-TrasactionRelayerAPI` to `...-TransactionRelayServiceAPI` maybe.

LSPs/LSP-15-TransactionRelayerAPI.md
Hugoo: it is because then we can use isValidSignature from the backend to check if that signature was a "true" signature.

frozeman (approved):

PR #135, opened 2022-11-02
Update LSP0 to include more specifications

YamenMerhi:
- Updating the LSP0 Specification to describe the behavior wanted for every LSP0 compliant contract.

> Just review the specification section, not the other sections :)

### DIFF
--- a/LSPs/LSP-0-ERC725Account.md
+++ b/LSPs/LSP-0-ERC725Account.md
@@ -70,7 +70,7 @@ MUST emit a [`ValueReceived`](#valuereceived) event when receiving native token.
 fallback() external payable;
 ```
 
-This function is part of the LSP17 specification, with additional requirements as follows:
+This function is part of the [LSP17]((./LSP-17-ContractExtension.md)) specification, with additional requirements as follows:
 
 - MUST be payable.
 - MUST emit a [`ValueReceived`](#valuereceived) event if value was present.
@@ -165,7 +165,7 @@ function setData(bytes32 dataKey, bytes memory dataValue) external;
 ```
 This function is part of the [ERC725Y](https://github.com/ERC725Alliance/ERC725/blob/develop/docs/ERC-725.md#setdata) specification, with additional requirements as follows:
 
-- MUST emit only the first 256 bytes of the dataValue in the DataChanged Event.
+- MUST emit only the first 256 bytes of the dataValue parameter in the [DataChanged](https://github.com/ERC725Alliance/ERC725/blob/develop/docs/ERC-725.md#datachanged) Event.
 
 
 #### setData (Array)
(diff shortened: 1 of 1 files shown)


LSPs/LSP-0-ERC725Account.md
```diff
@@ -41,65 +41,196 @@ This allows us to:
...
+
+
+#### fallback
+
+```solidity
+fallback() external payable;
 ```
 
-Check [LSP1-UniversalReceiver] and [LSP2-ERC725YJSONSchema] for more information.
+This function is part of the [LSP17] specification, with additional requirements as follows:
```
frozeman, line 73: link to the spec is important later

LSPs/LSP-0-ERC725Account.md
```diff
@@ -41,65 +41,196 @@ This allows us to:
...
-- [ERC1271](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1271.md#specification)
-- [LSP1](./LSP-1-UniversalReceiver.md#specification)
-- [LSP14](./LSP-14-Ownable2Step.md#specification)
+#### pendingOwner
+
+```solidity
+function pendingOwner() external view returns (address);
+```
+
+This function is part of the [LSP14] specification.
```
frozeman, line 94: Same on these.

YamenMerhi:
@frozeman Link are added, you can view the file and try them.
The actual link was added at the end of the file to avoid duplication and to be cleaner, but it's fully working :)

LSPs/LSP-0-ERC725Account.md
```diff
@@ -41,65 +41,196 @@ This allows us to:
...
+```solidity
+function universalReceiver(bytes32 typeId, bytes memory receivedData) external payable returns (bytes memory);
 ```
 
-The fallback function allows for receiving native tokens, as well as arbitrary calldata. The reasoning is that it allows for Graffiti with transactions, or protocols to be built for offchain parsing.
+This function is part of the [LSP1] specification, with additional requirements as follows:
+
+- MUST emit a [`ValueReceived`] event before external calls if the function receives native tokens.
+
+- If an `address` is stored under the data key attached below and and this address is a contract that supports the [LSP1UniversalReceiver interface id], forwards the call to the [`universalReceiver(bytes32,bytes)`] function on the address retreived appended with the caller address without padding (bytes20) and the uint256 msg.value sent (bytes32) to the `universalReceiver(bytes32,bytes)` function of the LSP0 contract. If there is no address stored under this data key, execution continues normally.
```
frozeman, line 205: why `address without padding (bytes20) and` ?? it `with padding`

Maybe this could be two sentences with a  new pin in the middle, so its not overseen, and maybe a bit simpler explained? I am not sure if people get how that bytes are added to `msg.data` (maybe mention `msg.data`), like:


```md

If an `address` is stored under the data key attached below and and this address is a contract that supports the [LSP1UniversalReceiver interface id], forwards the call to the [`universalReceiver(bytes32,bytes)`] function on the address retreived.

The `msg.data` is appended with the caller address as bytes20 and the `msg.value` received as uint256 bytes32 before calling the external contract, allowing the receiving contract to know the initial caller and the value sent.
```

frozeman (changes requested):
see comments

PR #141, opened 2022-11-23
LSP18 Royalties

lykhonis:


Hugoo:
Hey cool @lykhonis !
Not sure about the number tho as @CallumGrindle and I have a draft PR for LSP15 too https://github.com/lukso-network/LIPs/pull/133 😊😅🤭

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,293 @@
...
+This standard describes a set of [ERC725Y](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-725.md) data key values to store royalties recipient addresses and corresponding percentages in a [ERC725Y](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-725.md) smart contract.
+
+## Abstract
+This data key value standard describes a set of data keys that can be added to an [ERC725Y](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-725.md) smart contract to describe royalties:
+
+- `LSP18Royalties[]` is an [LSP2 array](./LSP-2-ERC725YJSONSchema.md) of royalties recipient addresses.
+- `LSP18RoyaltiesMap` is a dynamic address mapping, which contains:
+  - royaties percentage allocation per recipient address.
+  - and the index in the `LSP18Royalties[]` array.
+- `LSP18RoyaltiesEnforceAtLoss` is a boolean to enforce royalties whenever the NFT is sold at loss.
```
CJ42, line 22: I am not sure about this data key, as it is related to NFTs and seems to be relevant to a specific application logic.
But there is definitely something relevant around the "enforcing of payment".

Maybe it could be renamed like this? :)

```suggestion
- `LSP18RoyaltiesEnforcePayment` is a boolean to enforce royalties whenever the NFT is sold at loss.
```

TBD for the description.

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,293 @@
...
+
+The data key `LSP18RoyaltiesMap` exists so that smart contracts can detect if an address is present in the array (e.g. as done in the  [LSP1-UniversalReceiverDelegate](./LSP-1-UniversalReceiver.md)).
+
+## Motivation
+This standard allows to create a decentralised asset royalties allocations by a smart contract.
+
+## Specification
+Every contract that supports the LSP18DigitalAssetMetadata SHOULD have the following data keys:
```
CJ42, line 30:
```suggestion
Every contract that supports the LSP18Royalties SHOULD have the following data keys:
```

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,293 @@
...
+    "key": "0xe010b84f60176d2f466b495da33173c1a008950ed855b4b92282e356d009bd76",
+    "keyType": "Array",
+    "valueType": "address",
+    "valueContent": "Address"
+}
+```
+
+For more info about how to access each index of the `LSP18Royalties[]` array, see [ERC725Y JSON Schema > `keyType`: `Array`](https://github.com/lukso-network/LIPs/blob/master/LSPs/LSP-2-ERC725YJSONSchema.md#array)
+
+#### LSP18RoyaltiesMap
```
CJ42, line 52: I really like this data key.

The concept of points in particular allows some very good precision of the percentage. And it is also common language around financial subjects to say "your portfolio increased by x points, or decreased by y points".

Really like this whole section 😃

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,293 @@
...
+
+References royalties amounts of each recipient addresses.
+
+The data value MUST be constructed as follows: `bytes8(indexNumber) + bytes2(type) + bytesN`. Where:
+- `indexNumber` = the index in the [`LSP18Royalties[]` Array](#lsp18royalties).
+- `type` = the type of a payload:
```
CJ42, line 58: What does "payload" refer to here?

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,293 @@
...
+        "name": "LSP18RoyaltiesEnforceAtLoss",
+        "key": "0x9543ef82e4010b6d1749c01487661e9bc5109a255c17b906cc3bc37fa994874e",
+        "keyType": "Singleton",
+        "valueType": "boolean",
+        "valueContent": "Boolean"
+    }
+]
+```
+
+```solidity
```
CJ42, line 121: We are trying to avoid having custom Solidity code in the LIP specs, because we want to emphasize that the LIP specs are language agnostics. So the implementation could be written in any smart contract language (Solidity, Vyper, Fe, Rust, inK, etc...).

Maybe you can add a link instead to your repo? 🙂

CJ42 (commented):
Thank you @lykhonis for this LIP proposal.

It seems very interesting to me and a very relevant Standard proposal with a lot of potential use cases 🚀 

I have added some review comments.

Let's wait for @frozeman @YamenMerhi @skimaharvey and @b00ste review for more insights.

LSPs/LSP-18-Royalties.md
lykhonis: Correct. For now just these 2 types, this can be extended later for example for a delegated calculation such as type could be an address of an interface (contract) to call calculation logic etc.

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,122 @@
...
+    - `1` = fixed amount in `LYX`
+        - `bytes32(amount)` = wei amount
+
+
+```json
+{
+    "name": "LSP18RoyaltiesMap:<address>",
+    "key": "0xe5ea91873c26eca6dec90000<address>",
+    "keyType": "Mapping",
+    "valueType": "(bytes8,bytes2,bytes4)",
```
YamenMerhi, line 70: How the last value type is bytes4 where in case of payload 1 the amount will be bytes32 ? Am I missing something ?
Also is there a specific intention having the type as bytes2 and the amount as bytes4/bytes32 ?

HarmonyONELOVER (approved):

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,122 @@
...
+}
+```
+
+For more info about how to access each index of the `LSP18Royalties[]` array, see [ERC725Y JSON Schema > `keyType`: `Array`](https://github.com/lukso-network/LIPs/blob/master/LSPs/LSP-2-ERC725YJSONSchema.md#array)
+
+#### LSP18RoyaltiesMap
+
+References royalties amounts of each recipient addresses.
+
+The data value MUST be constructed as follows: `bytes8(indexNumber) + bytes2(type) + bytesN`. Where:
```
CJ42, line 56: I would suggest that the first value in the tuple is a `uint128` (16 bytes long), the same type + bytes length as the index for the `LSP18Royalties[]` Array. 

This will prevent indexes being truncated for index larger than `max(uint64)`.

**Important: I would mark it as a `uint128` ✅ , not as a `bytesN` because of the padding rules (left padded for uint, instead of right padded for bytes)**. The same as it is now with `LSP5ReceivedAssets[]` and `LSP5ReceivedAssetsMap:<address>` when we refactored.

image


```suggestion
The data value MUST be constructed as follows: `uint128(indexNumber) + bytes2(type) + bytesN`. Where:
```

LSPs/LSP-18-Royalties.md
CJ42, line 70: suggested change, so that it reflects properly the type for the Array `index`

```suggestion
    "valueType": "(uint128,bytes2,bytes4)",
```

CJ42 (changes requested):

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,130 @@
...
+    "key": "0xe6156b5402ae58a1fbf00000<address>",
+    "keyType": "Mapping",
+    "valueType": "(bytes4,uint128,uint8,uint256)",
+    "valueContent": "(Bytes4,Number,Number,Number)"
+}
+```
+
+#### LSP18RoyaltiesEnforcePayment
+
+A boolean when `true` to indicate enforcement of royalties even if a NFT is sold at a loss. By default this is `false` and a marktetplace may not enforce royalties.
```
skimaharvey, line 86: should we have a datakey that store the latest sold value then?

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,130 @@
...
+### ERC725Y Data Keys
+
+#### LSP18RoyaltiesRecipients[]
+
+An array of royalties recipients (see [LSP-0-ERC725Account](./LSP-0-ERC725Account.md)).
+
+The data value MUST be constructed as follows: `address(recipient) + uint8(valueType) + uint256(value)`. Where:
+- `recipient` = address of a recipient to receive royalties
+- `valueType` = type of royalties:
+    - `0` = a percentage in points where `100_000` is basis. e.g. `15%` is `15_000` points, and `1.5%` is `1_500` points.
```
skimaharvey, line 46: even though it is what make the most sense imo you will never be able to enforce royalties with percentage.
I can always transfer it for 0 value to a "middle" man smart contract that will have some logic that prevent paying royalties

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,130 @@
...
+    - `1` = a LYX amount in wei.
```
skimaharvey, line 47: what will happen if I transfer it for 0 and enforcement is true? who will pay for royalties if `msg.value == 0`?

skimaharvey (commented):
Looks good. 
The only thing is that we might face some real challenges with the notion of "enforcing" royalties. 
Whether we use a percentage or  LYX amount, they can always be bypassed imo.

skimaharvey:
> > Looks good. The only thing is that we might face some real challenges with the notion of "enforcing" royalties. Whether we use a percentage or LYX amount, they can always be bypassed imo.
> 
> Honestly not sure how to start answering about "enforcement" part. This is a standard describing metadata.

Ah ok sorry thought it was a standard that intended to be implemented

LSPs/LSP-18-Royalties.md
skimaharvey: well my comment is under the `  - `1` = a LYX amount in wei.` I already gave my comment about the percentage enforcement that was not implementable

LSPs/LSP-18-Royalties.md
jakeprins: > even though it is what make the most sense imo you will never be able to enforce royalties with percentage. I can always transfer it for 0 value to a "middle" man smart contract that will have some logic that prevent paying royalties

This standard is not about always enforcing royalties but about setting royalties settings on the NFT contract, so marketplaces (like universal.page) can enforce them.

jakeprins (approved):

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,130 @@
...
+}
+```
+
+#### LSP18RoyaltiesEnforcePayment
+
+A boolean when `true` to indicate enforcement of royalties even if a NFT is sold at a loss. By default this is `false` and a marktetplace may not enforce royalties.
+
+```json
+{
+    "name": "LSP18RoyaltiesEnforcePayment",
```
CallumGrindle, line 90: The confusion about enforcement comes from this `LSP18RoyaltiesEnforcePayment` key. Enforcement of royalties is tricky because even if your marketplace is enforcing royalties, I can just go to another marketplace which does not enforce royalties so I can buy my NFT for cheaper. 

It's not totally clear how this can be 'standardised' if people can opt out of paying. I.e. whats the point in standardising a metadata key for enforcing royalty payments, if royalty payments cannot be enforced? 

I would suggest at least addressing this in the motivation section of the standard, or find a way this can be implemented at a smart contract level

jakeprins: @CallumGrindle, don't you think having some sort of standard around royalties settings is good? So instead of each platform doing its own thing, these settings can be stored on-chain so the marketplaces could act on it?

skimaharvey: Alright, got it, so it's standard tailored for centralized marketplaces. 
How utterly refreshing! Because that was the mysterious puzzle piece I was losing sleep over 😄

jakeprins: > Alright, got it, so it's standard tailored for centralized marketplaces. How utterly refreshing! Because that was the mysterious puzzle piece I was losing sleep over 😄

Great, let's close the PR. I didn't expect it to take over half a year to come to that conclusion (although I could have known better).

skimaharvey: > @CallumGrindle, don't you think having some sort of standard around royalties settings is good? So instead of each platform doing its own thing, these settings can be stored on-chain so the marketplaces could act on it?

It does make sense in the meantime we find a way to enforce them at the SC level

CallumGrindle: @lykhonis @jakeprins indeed that makes sense. The question is more about whether it SHOULD be a standard about enforcing royalty payments given those are the keys that are being suggested. 

The issue I see is the following:
- A user creates an NFT collection
- The user selects to enforce royalties 
- There is a marketplace which decides to not enforce royalties
- People start buying the NFTs from this marketplace
- The user gets no royalty payments decide setting that it should be enforced

Its still a useful key for the marketplaces which do adhere to the standard properly, but my worry is that by having 1 marketplace which does not implement the standard, it becomes a problem for the marketplaces which do and the royalty recipients.

Addressing this problem somewhere in the motivation, and stating that it should be handled elsewhere would not hurt

samuel-videau: If I can add my neutral input on this @CallumGrindle @skimaharvey @lykhonis @jakeprins. 
I think it does not matter if some marketplaces decide to use it or not. What matters is that marketplaces that want royalties (as it's hardly enforcable), can do it the same way.
Because if not, something we risk to see is having a lot of different keys for the same thing:
`Markerplace1Royalties`, `Marketplace2Royalties`, `Marketplace3Royalties`, and that would be super annoying for NFT collections creators/developers (without even talking about storage waste)

skimaharvey: Yes, it is useful but would be better at the smart contract level for the reason @CallumGrindle mentioned... 
Like if I had to sell an NFT I would not shut myself in the foot and would use a marketplace that doesn't enforce them as it increases my chances of selling my NFT. Let's see for how long marketplaces enforce them when they are getting bypassed because they are enforcing something while others don't. 
Anyway, if the marketplaces then we will need a standard for it.

frozeman:
@lykhonis the focus on the team is the network and up launch right now. Additional standards have less importance internally yet. 

Tho this doesn’t mean you need to close this. Standard debates can happen over months and years and closing and implement as a single project is likely not leading to adoption. 

I personal think the buyable nft standard is the better approach to royalties. But I had no time to look at this proposal in the mean time. 

Also please note. Standard adoption is not determined by a LSP being added to the repo, but by people using it. 
I would suggest you reopen it and start using it as you see fit. And I’ll give my option after the network launch up process is over and I got more mental space for additional ideas. 
(but like I said standards don’t need the lukso team approval, or mine to be used by anyone)

LSPs/LSP-18-Royalties.md
CJ42, line 46: One thing I am not sure about is that the values `uint8(valueType) + uint256(value)` are present in both:

- the `LSP18RoyaltiesRecipients[]` Array data key
- the `LSP18RoyaltiesRecipientsMap:<address>` data key.

This could be a problem, as both entries have to be in sync. For instance, what happen if royalty settings get updated and let's say:
- in the Array data key you have `valueType == 0` (a percentage for the type of royalty) (assumed to be the correct setting after update) ✅ 
- in the Map data key you have `valueType == 1` (a fixed LYX amount) ❌ (old setting that was not updated).

I would suggest 3 options:

_option 1: keep it as it is, the settings are in both the Array and Map data key, but the standard specify that **they MUST always be in sync (so the implementation code MUST ensure that)**_

_option 2: remove these two values from the Array data key (the Array data key contains only the Royalty recipient's address), and only the Map contains the royalty settings (same as it is done in LSP5, LSP10 and LSP12)_

image

LSPs/LSP-18-Royalties.md
CJ42, line 86:
```suggestion
A boolean when `true` to indicate enforcement of royalties even if a NFT is sold at a loss. By default, this is `false`, and a marketplace may not enforce royalties.
```

CJ42 (commented):
@jakeprins @lykhonis I am glad to see that this PR was re-opened.

I have left some review comments and would like to merge and accept this proposal, as it seems that everyone agrees on having a standard around Royalties.

There are two main points I would like to point out in particular

**1. having the settings in both the Array + Map seems not necessary**

At least from my viewpoint. I have added in the comments below. I would suggest that the Map data key stores the actual settings, while the array entries stores only the address of the royalty recipients.
But I might lack insights and I don't have the same perspective (seems you have been working on that for a while now).

If you think it's necessary to have the settings in both entries, it would be great if you could explain the reason why it in the standard, so that readers can understand the reason behind this design decision.

**2. About the royalty enforcement debate**

It seems that the royalty enforcement part is what is bringing the most debate in the proposal. I would suggest the following to help everyone reach consensus.

--> add a Security Consideration section (similar to LSP17, see screenshot below), where you mentioned what you said that _"this standard is not about royalty enforcement, but purely for the purpose of setting royalties configurations in a contract"_. You can as well mention the security aspects brought by @skimaharvey and @CallumGrindle in this section, so that developers who will adopt the standards are aware of these points, and understand that **they should use this standard purely for storing royalty settings, and not for enforcing them. Royalty enforcement should be done at the application level (the marketplace itself)**.

image


@frozeman I guess we can always use the Buyable NFT standard in the future as an alternative way around Royalties, so that people have options.

samuel-videau:
After re-reading the standard, I'm just wondering the need of 3 different keys for that, I think one would be more than enough.

Firstly, I think `LSP18RoyaltiesEnforcePayment` doesn't make sense, as if the other royalty keys are filled, then it means you want marketplaces to enforce royalties. I won't create a NFT, add royalties on it but then say "no do not apply those royalties", if that value is on false, then why adding royalties to the contract ? So basically, marketplaces should assume that if the royalty keys are filled, then creators want them to be applied.

Secondly, the two other keys are really redondant. Is the `tandardInterfaceId` really important?
Let's say it is, why not having only the `LSP18RoyaltiesRecipients[]` key with the following format:
```
{
    "name": "LSP18RoyaltiesRecipients[]",
    "key": "0xfdd4e98ba62fdcf79cfde4cfe031a71195ae21ff3d0e29f79db24f2fe9ceb59b",
    "keyType": "Array",
    "valueType": "(address,bytes4,uint16,uint32)",
    "valueContent": "(Address,Bytes4,Number,Number)"
}
```

Using this key only you would have all the info you need to add royalties (or am I missing something?). And the value would be packed in less than 32 bytes

CJ42:
@samuel-videau The Map data key is needed because otherwise to get the royalty settings of an address, you would have to loop through the entire array in storage. Which can be big a uses a lot of gas.

samuel-videau:
> @samuel-videau The Map data key is needed because otherwise to get the royalty settings of an address, you would have to loop through the entire array in storage. Which can be big a uses a lot of gas.
> 
> You could put everything in one data key and allow it. But the standard needs to specify this aspect and make developers aware of it.

But the thing is that here you still need to loop through the array to know the addresses that should have royalties anyway

CJ42:
That's actually a pretty good point. I agree about this.

skimaharvey:
> 3. LSP18RoyaltiesEnforcePayment is needed as it instructs a marketplace to enforce royalties regardless whether a user is selling it at a profit or at a loss. If this key is set to false, and user is selling at a loss, a marketplace should not enforce royalties and allow sale without them.

How will the concept of "selling at a loss" will work cross-marketplaces since the value is not part of any events? 
For example, I buy my NFT on opensea and sell it on looksrare. How will Looksrare know about the price I purchased my nft for? Should looksrare rely on opensea api 👀

LSPs/LSP-18-Royalties.md
FamilyNFT: @lykhonis @jakeprins, I think this would be a far more impactful standard if we we're to enforce royalties at the core level. We should be putting the creators at the forefront as we're The New Creative Economy. Simply creating a standard for marketplaces will only repeat what has happened in the past where different marketplaces will try get market share by removing royalties, making creators lose out at the end.....

If you guys truly value creators like you said, we should be thinking at the fundamental levels. This will also not just benefit creators but Universal Page aswell, since you'll be removing upcoming competition that tries to compete using "royalties" as their edge.

Any chance you guys can make the necessary changes to make this enforceable at the smart contract level please @skimaharvey @CallumGrindle. We would love to implement this in Family

frozeman:
@lykhonis sorry for the long wait, we now had a proper look internally and the standard looks great, as separating issuers and royalties makes sense.

We would have a few points that I will comment on the doc.

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,122 @@
+---
+lip: 18
+title: Royalties
+author: Volodymyr Lykhonis <vlad@universal.page>, Jake Prins <jake@universal.page> 
+discussions-to: https://discord.gg/E2rJPP4
+status: Draft
+type: LSP
+created: 2022-11-23
+requires: LSP2, LSP4
```
CJ42, line 9: Why is LSP4 required here? I think it is not, we can remove this.

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,122 @@
...
+## Specification
+Every contract that supports the LSP18Royalties SHOULD have the following data keys:
+
+### ERC725Y Data Keys
+
+#### LSP18RoyaltiesRecipients[]
+
+An array of royalties recipients (see [LSP-0-ERC725Account](./LSP-0-ERC725Account.md)).
+
+The data value MUST be constructed as follows: `address(recipient) + uint32(value)`. Where:
```
CJ42, line 41: What is the rationale for choosing `uint32` as valueType?

frozeman:
After additional thinking.
1. doing automatic push payments of NFT sales will require the payout smart contract to iterate over the array keys, which can be very costly.
2. Using `Array` keyType was probably chosen so that it supports many royalty receivers.


I see two scenarios
1. a few royalty receivers. -> Then an compact array as `valueContent` is better and keyType `Singleton`. As this is one call, and easy to iterate over.
2. have many royalty receivers -> Its better to point to a LSP7 contract, where each balance references the percentage e.g. 1000 tokens available, and balance means a percentage of this.

For small amount of royalty receivers its easy todo a "push" payment distribution.  (just look up the royalties array)

For large royalty receivers group, you need to use a "pull" contract. e.g.
on NFT purchase, the royalties get put into a claim contract and based on the LSP7 royalty balances, users can withdraw their part.

To make both of these use cases possible i would suggest:

**Small number of royalties receivers**

```js
{
    "name": "LSP18InternalRoyaltiesRecipients",
    "key": "0xxxxxxxxx",
    "keyType": "Singleton",
    "valueType": "(address,uint32)[CompactBytesArray]",
    "valueContent": "(Address,Number)"
}
```

[Compact byte array](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#bytescompactbytesarray) allows for more simple encoded arrays.

basically you get back in one single smart contract call `[(reciver1, percentage), (reciver2, percentage), ...]`

**Large number of royalties receivers**

```js
{
    "name": "LSP18ExternalRoyaltiesRecipients",
    "key": "0xxxxxxxxx",
    "keyType": "Singleton",
    "valueType": "(bytes4,uint32,address)",
    "valueContent": "(Bytes4,Number,Address)"
}
```

- `Bytes4` is the interfaceId, of the linked contract
- `Number` is the percentage of royalties that get distributed to all holders of the linked contract (to be used by the distribution contract, which could also be the same as the linkedin contract )
- `Address` The address is a LSP7 or any other contract, that contains the royalty distribution

lykhonis:
> After additional thinking.
> 
> 1. doing automatic push payments of NFT sales will require the payout smart contract to iterate over the array keys, which can be very costly.
> 2. Using `Array` keyType was probably chosen so that it supports many royalty receivers.
> 
> I see two scenarios
> 
> 1. a few royalty receivers. -> Then an compact array as `valueContent` is better and keyType `Singleton`. As this is one call, and easy to iterate over.
> 2. have many royalty receivers -> Its better to point to a LSP7 contract, where each balance references the percentage e.g. 1000 tokens available, and balance means a percentage of this.
> 
> For small amount of royalty receivers its easy todo a "push" payment distribution. (just look up the royalties array)
> 
> For large royalty receivers group, you need to use a "pull" contract. e.g. on NFT purchase, the royalties get put into a claim contract and based on the LSP7 royalty balances, users can withdraw their part.
> 
> To make both of these use cases possible i would suggest:
> 
> ```js
> {
>     "name": "LSP18RoyaltiesRecipients",
>     "key": "0xxxxxxxxx",
>     "keyType": "Singleton",
>     "valueType": "(bytes4,address,uint32)[CompactBytesArray]",
>     "valueContent": "(Bytes4,Number,Address)"
> }
> ```
> 
> [Compact byte array](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#bytescompactbytesarray) allows for more simple encoded arrays.
> 
> basically you get back in one single smart contract call `[(reciver1, percentage), (reciver2, percentage), ...]`
> 
> **Small number of royalties receivers** Simply put an array of UPs or EOAs in there
> 
> **Large number of royalties receivers** Just put the address of an LSP7 contract with `claim/withdraw` function, that distributes based on tokens held in the LSP7.
> 
> you can even have a combination of both now.
> 
> * `Bytes4` is the interfaceId, of the linked contract (TBD if that's necessary)
> * `Number` is the percentage of royalties that get distributed to the address
> * `Address` The address of receiver(), which could also be a LSP7 royalty claim contract(s) or any other contract, that contains the royalty distribution system
> 
> **Benefits**
> 
> * allows for cheaper sort contract calls for small amount of royalties receivers
> * allows for very large group where royalties can even be changing over time, as its could be a LSP7
> 
> side note: i would remove the dependency on LSP4, that's necessary
> 
> -> Concerning enforcement data key, what's your thinking here? @lykhonis

I think indeed this is a good change. We can start slowly with a small number of receivers (most likely the most common case anyway) as a single key/value.

Enforcing part was more marketplace dependent feature. Rational was for a creator to opt-in to not charge royalties if a NFT is sold at a loss. If I to buy nft at $10 and sell at $5, I will not pay 10% royalties thus loosing even more.

jakeprins:
> @frozeman I guess we can always use the Buyable NFT standard in the future as an alternative way around Royalties, so that people have options.

@CJ42 For the buyable NFT standard. Will there be a way to enforce a fee for the curator? Enforceable royalties for NFT creators are great, but without enforceable royalties/fees for platform creators (like Universal Page), there is no incentive to create the products to discover and trade these assets. This lack of incentive could hurt the NFT creators (and overall ecosystem) even more so I'm interested to hear what you think about this.

lykhonis:
We addressed compact array solution in the LSP and implemented it locally in solidity and typescript. I believe it is a good baseline and start.

frozeman:
@jakeprins sorry for the late response. In the future feel free to ping me in DM on discord. So I see it and answer faster.

In my opinion, the “en force” part should be in the marketplace itself. Not the standard. In my opinion, it should be by the fault always requiring royalties. But if it’s on the marketplace level, if anyway depends on the marketplace implementation . If we later add the ability to buy and sell the NFT directly over the nft smart contract, then royalties will be always enforced, or, however, the order book contract attached to the NFT decides to do it. 
The entries here in the NFT are purely information as of now, and it depends on the marketplace or future standards to use it in whatever way they want anyway.

jakeprins:
> @jakeprins sorry for the late response. In the future feel free to ping me in DM on discord. So I see it and answer faster.
> 
> In my opinion, the “en force” part should be in the marketplace itself. Not the standard. In my opinion, it should be by the fault always requiring royalties. But if it’s on the marketplace level, if anyway depends on the marketplace implementation . If we later add the ability to buy and sell the NFT directly over the nft smart contract, then royalties will be always enforced, or, however, the order book contract attached to the NFT decides to do it. The entries here in the NFT are purely information as of now, and it depends on the marketplace or future standards to use it in whatever way they want anyway.

@frozeman Yes indeed. However, the original question was more about the "buyable" NFT standard and how it could enforce "royalties" for the creators of NFT platforms (known as platform fees). I think it's important to incentivize the creation and maintenance of these platforms, especially for applications that require a lot of work and money to build and run. Otherwise, the "royalty" problem would simply be shifted from NFT creators to app creators, which I think would not result in a healthy ecosystem in the long run (assuming it will gets widely adopted). At the very least, it would force app creators to adopt a different business model. But if we look at "fee-less" marketplaces, this can quickly become questionable. For example, a lot of platforms launch (mostly useless) tokens which the community often pays the price for. Or you could go the Web2 way of filling everything with ads and tracking scripts (also not that cool). If there's one thing people should have learned from Web2, it's that _if you're not paying for the product, you are the product_. Therefore I believe the "fee" model is (just like creator royalties) a better option for app creators. Anyway, this is just my personal opinion, and I'm obviously biased as I want Universal Page to succeed and stay around for a long time.

FamilyNFT:
@jakeprins from my perspective the buyable standard doesn't need to enforce royalties for curators/middleman since this is already built in to the overall infrastructure of blockchains. What I mean by this is that end users still needs to pay gas/fees to be able to use the blockchain therefore as a curator/platform they inherently benefit from this feature. So they either abstract it away from end users via ads or simply pay for user's gas by taking a fee. And in my humble opinion, curators should play the game of AND not OR for the best business results ;)

CJ42:
> @jakeprins sorry for the late response. In the future feel free to ping me in DM on discord. So I see it and answer faster.
> 
> In my opinion, the “en force” part should be in the marketplace itself. Not the standard. In my opinion, it should be by the fault always requiring royalties. But if it’s on the marketplace level, if anyway depends on the marketplace implementation . If we later add the ability to buy and sell the NFT directly over the nft smart contract, then royalties will be always enforced, or, however, the order book contract attached to the NFT decides to do it. The entries here in the NFT are purely information as of now, and it depends on the marketplace or future standards to use it in whatever way they want anyway.

I also see this enforcable part in the Marketplace itself, so that different market places can come up with different solutions for enforcing royalties, which would create different business models that NFT creators could choose from.

The Marketplace could create a LSP17 Extension contract in the form of the order book that contains the royalty enforcement logic (that the Marketplace itself has defined). This way, we can differentiate between the storage layer (the royalty distribution settings, percentage, etc...), and the logic layer (in the LSP17 extension itself). An NFT collection could then decide to move overtime from one royalty distribution medium to another, if it finds a new one more efficient, with additional benefits.

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,97 @@
...
+{
+    "name": "LSP18RoyaltiesRecipients",
+    "key": "0xc0569ca6c9180acc2c3590f36330a36ae19015a19f4e85c28a7631e3317e6b9d",
+    "keyType": "Singleton",
+    "valueType": "(bytes4,address,uint32)[CompactBytesArray]",
+    "valueContent": "(Bytes4,Address,Number)"
+}
+```
+
+A compact byte array allows optinally to store additional fields if needed. Required fields are: `interfaceId` and `recipient`. The `points` field is optional and if not provided it is assumed to be `0` points.
```
Hugoo, line 55:
```suggestion
A compact byte array allows optionally to store additional fields if needed. Required fields are: `interfaceId` and `recipient`. The `points` field is optional and if not provided it is assumed to be `0` points.
```

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,97 @@
...
+Every contract that supports the LSP18Royalties MUST have the following data keys:
+
+### ERC725Y Data Keys
+
+#### LSP18RoyaltiesRecipients
+
+An array of royalties recipients and corresponding percentages.
+
+The data value MUST be a [LSP2 BytesCompactArray](./LSP-2-ERC725YJSONSchema.md#bytescompactbytesarray) which contains a list of royalties recipients. Each royalties recipient is a tuple of:
+- `interfaceId` = an interface identifing a recipient of royalties. If the interface is not known, it is assumed to be `0xffffffff`.
```
Hugoo, line 41: Does it mean that if it is set to `0xffffffff`, then the interfaceId check is skipped? Which means royalties will be sent to any kind of contract or EOA defined as `recipient`?

LSPs/LSP-18-Royalties.md
```diff
@@ -0,0 +1,97 @@
...
+    "key": "0xc0569ca6c9180acc2c3590f36330a36ae19015a19f4e85c28a7631e3317e6b9d",
+    "keyType": "Singleton",
+    "valueType": "(bytes4,address,uint32)[CompactBytesArray]",
+    "valueContent": "(Bytes4,Address,Number)"
+}
+```
+
+A compact byte array allows optionally to store additional fields if needed. Required fields are: `interfaceId` and `recipient`. The `points` field is optional and if not provided it is assumed to be `0` points.
+
+#### LSP18RoyaltiesEnforcePayment
```
frozeman, line 57: I am not sure about this, because its not clear what this is good for, if interfaces/websites/trading contracts anyway make the decision to enfore or not by themselves.

CJ42 (approved):

CJ42:
Super! First community standard 🎉

PR #142, opened 2022-11-24
Add first draft of LSP17ContractExtension

YamenMerhi:
- First draft of LSP17ContractExtension

### DIFF
--- a/LSPs/LSP-17-ContractExtension.md
+++ b/LSPs/LSP-17-ContractExtension.md
@@ -16,13 +16,17 @@ This standard describes a way to extend contract's functionalities even after de
  
 ## Abstract
 
-This proposal defines two types of contracts, the extendable and the extensions contract. An extendable contract is a contract that in case of being called with a function that does not exist, it calls an extension contract ,through the fallback function, with the calldata received appended with the `msg.sender` and `msg.value` of the extendable contract as extra calldata.
+This proposal defines two types of contracts:
+- an extendable contract which functionalities are extended.
+- an extension contract that extend the functionalities of the extendable contract.
 
-The extendable contract should map function selectors (bytes4) to extensions (address) that implement these functions being called.
+When the extendable contract receives a call for a function not implemented in its public interface, it can forward this call to an extension contract. This forwarding occurs through the fallback function. The extension contract can then receive the calldata of the initial message call, appended with the `msg.sender` and the `msg.value` initially received.
+
+The extendable contract should map function selectors (bytes4) to extension contract addresses that implement these functions being called.
 
 ## Motivation
 
-After deploying a contract, there is no possible way to integrate native functions into the bytecode of the deployed contract. This represents a limitation that smart contract have specially with smart contract based account that could evolve by time and needs to have specific function to support future usecases and standards.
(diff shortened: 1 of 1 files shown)


LSPs/LSP-17-ContractExtension.md
```diff
@@ -0,0 +1,190 @@
...
+
+If the extendable contract does not support ERC725Y, the [ExtensionChanged] event MUST be emitted whenever an extension is added/changed/removed.
+
+### fallback Function
+
+Here is an illustrative example of how the extendable contract's fallback function might be implemented:
+
+```solidity
+
+fallback() external payable {
```
frozeman, line 75: I think this should be referenced as implementation examples but not in the spec. The reason why standard specs shouldn't have code like this, as they should be ideally smart contract language independent.

LSPs/LSP-17-ContractExtension.md
```diff
@@ -0,0 +1,190 @@
...
+- The `msg.sender` calling the extendable contract without any pad, MUST be 20 bytes.
+- The `msg.value` received to the extendable contract, MUST be 32 bytes.
+
+
+If the extendable contract supports [ERC725Y], the extension address MUST be stored under the data key attached below:
+
+```json
+{
+    "name": "LSP17Extension:<bytes4>",
+    "key": "0xcee78b4094da860110960000<bytes4>",
```
CJ42, line 56: I think this is confusing, because here we have:

cee78b4094da860110960000 (12 bytes) + `<bytes4> (4 bytes) = 16 bytes.

While the data key is 32 bytes long.

We should indicate that it is **left padded** + reference LSP2 for clarity.

https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#mapping

image

LSPs/LSP-17-ContractExtension.md
```diff
@@ -0,0 +1,190 @@
...
+    "keyType": "Mapping",
+    "valueType": "address",
+    "valueContent": "Address"
+}
+```
+
+The [DataChanged] event MUST be emitted whenever an extension is added/changed/removed.
+
+The LSP17ContractExtension do not enforce a specific way to store the extension address based on the bytes4 function selector. It could be an explicit mapping from function selectors to extensions, eg: `mapping(bytes4 => address)`.
```
CJ42, line 65: _Suggested changes_

```suggestion
The LSP17ContractExtension does not enforce a specific way to store the extension address based on the bytes4 function selector. As an example, a mapping from function selectors to extensions, (eg: `mapping(bytes4 => address)`) can be used, but any other data structure can be used to map function selector to extension contract addresses.
```

LSPs/LSP-17-ContractExtension.md
```diff
@@ -0,0 +1,190 @@
...
+
+MUST be emitted when an extension is added/changed/removed.
+
+> In case the extendable contract supports [ERC725Y], there is no need to emit the ExtensionChanged event.
+
+
+
+## LSP17Extension Specification
+
+The key words “MUST”, “MUST NOT”, “REQUIRED”, “SHALL”, “SHALL NOT”, “SHOULD”, “SHOULD NOT”, “RECOMMENDED”, “MAY”, and “OPTIONAL” in this document are to be interpreted as described in [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
```
CJ42, line 136: It seems strange to me to have this content duplicated twice.

CJ42 (commented):
added review comments

frozeman (approved):

PR #143, opened 2022-11-27
Update LSP8 to include value 4 on enum LSP8TokenIdType with for hex encoded ASCII string

samuel-videau:
Might be relevant having a `string` value for the `TokenIdType` enum. It will then be possible to decode, in a standardized way, any token name. Each tokenId would be a hex encoded ASCII string.

The hex encoded string should be right padded, in order to use the mapping `LSP8TokenIdMetadata:TokenId` properly.
e.g. for a token name `super_token_name` we would have the following tokenId : `0x0000000000000000000000000000000073757065725f746f6b656e5f6e616d65`

Limitations of this token id type:
- ASCII string only
- 32 characters long string max
- As @fhildeb mentioned already, strings could be the same but with different capitalization (e.g. `My token` != `My Token`)


CJ42:
> 32 characters long string max

Why not using tokenId type `3` in this case? @samuel-videau 

The definition of token Id type `3` should probably be changed, to describe that it does not necessarily refer to a hashed value or a serial number. But could be also used as a string.

image

I think it is rather the description in LSP8 that is misleading.

samuel-videau:
> > 32 characters long string max
> 
> Why not using tokenId type `3` in this case? @samuel-videau
> 
> The definition of token Id type `3` should probably be changed, to describe that it does not necessarily refer to a hashed value or a serial number. But could be also used as a string.
> 
> image
> 
> I think it is rather the description in LSP8 that is misleading.

Hmm because, I guess, an app needs to be able to make the difference between a hashed value (cannot be "unhashed") and a hex value. 
How do my app knows if, on your `tokenIdType` 3 contract, the tokenIds are hashed values or just string ? 
I'll simply use the example of [Lookso](https://lookso.io), there if we want to translate a token transfer event, we'll need to know if a tokenId can be converted to string or not.

e.g. `keccak256('Hello')` = `06b3dfaec148fb1bb2b066f10ec285e7c9bf402ab32aa78a5d38e34566810cd2`
So if our app try to convert that to string, it will give `³ß®ÁHû²°fñÂ…çÉ¿@*³*§Š]8ãEfÒ`

I'd like having the input of @frozeman on that point

YamenMerhi:
I think the tokenId `hash or serial number` should be different from `string` as @samuel-videau mentioned.
Because it's 2 different things and interfaces should know how to handle the tokenId based on the type, so if we combine them it will create more confusion and more complexity on interfaces.

Also, `string` tokenId type seems logical for me as there could be some tokens that wants to simply make a token that have a standard name and then it could be the name as the tokenId.

skimaharvey:
Personally, I also think that if there is a use case for having `tokenId` as a string it should be differentiated from `hash` (3).
> * As @fhildeb mentioned already, strings could be the same but with different capitalization (e.g. `My token` != `My Token`)

To me, this is the biggest problem, when using strings it leaves room for interpretation. Some dApp may interpret your `tokenId` as "My token" and others will interpret it as "My Token" etc... Turkey != turkey

Also, I would be afraid that it encourages people to use descriptive `tokenId` over metadata which is standardized, helps ensuring consistency and avoid potential confusion

samuel-videau:
> Personally, I also think that if there is a use case for having `tokenId` as a string it should be differentiated from `hash` (3).
> 
> > * As @fhildeb mentioned already, strings could be the same but with different capitalization (e.g. `My token` != `My Token`)
> 
> To me, this is the biggest problem, when using strings it leaves room for interpretation. Some dApp may interpret your `tokenId` as "My token" and others will interpret it as "My Token" etc... Turkey != turkey
> 
> Also, I would be afraid that it encourages people to use descriptive `tokenId` over metadata which is standardized, helps ensuring consistency and avoid potential confusion

I agree it is the biggest problem. However, it's the developers responsibility to choose according to their requirements, and maybe to add some guards to the mint methods. Our responsibility is to inform, through our docs, of the pros and the cons of each solution.
I strongly believe that it should be available, as, for a lot of use cases, it might be really relevant.

CJ42 (approved):
@samuel-videau @skimaharvey we can merge this PR and I will integrate an extra row in the table on my PR  #148

skimaharvey (approved):

CJ42:
Closing as implemented in #148 ✅

Issue #136, opened 2022-11-03
[LSP8] Add explanation of hashed tokenId


lykhonis:
### Overview
LSP8 requires to set an explanation of what `tokenId` content means. The explanation is currently documented to be an integer of key [LSP8TokenIdType](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-8-IdentifiableDigitalAsset.md#lsp8tokenidtype):
- 1: `address` another contract
- 2: `uint256` a number, which may be an incrementing count where each minted token is assigned the next number
- 3: `bytes32` a hashed value (ie. serial number)

### Problem
In a case of `3` when `tokenId` represents a hashed value, dapps and other contracts may need to be able to retrieve actual value (prior to it's been hashed).

## Hashing Function

```json
{
    "name": "LSP8TokenIdHashFunction",
    "key": "0x200d7fc10771371795cceb6ad6d33d5959c47250486fb93f77f0c20088b90a97",
    "keyType": "Singleton",
    "valueType": "string",
    "valueContent": "String"
}
```

possible values:
- `keccak256(bytes)`

## Token Value

```json
{
    "name": "LSP8TokenIdValue:<bytes32>",
    "key": "0x60fcfd25909da10abc320000<bytes32>",
    "keyType": "Singleton",
    "valueType": "bytes",
    "valueContent": "Bytes"
}
```

If value is a string, it must be encoded int `utf8` and read as `Web3.utils.hexToUtf8(hexValue)`

frozeman:
But token if can be anything. From
A hash. A short utf8 encoded string. A number or a bit mask. 
I think naming 3 a hashed value is a mistake, it was never meant to be just a hash. 


maybe we should introduce more token types, rather than adding a second key (which would make it more expensive to deploy)

lykhonis:
I would give an example. For Universal Page we require a user provided name to track pages. Each name is then hashed keccak256. The hash becomes a token id. This works great, because we do not have a min nor maximum requirements for possible names (more than 32 bytes). Token type id `3` makes sense in this case. Only problem is that we cannot recover a page name from a hash, but we can generate token id from a name.

I still think this type `3` makes sense though. It adds additional flexibility on token ids, but needs few more details to recover the value.

CJ42:
I think it's rather the description in the LSP8 standard that is misleanding.

1: `address` = the tokenId that is minted is represented as its own ERC725Y contract.
[...]

PR #148, opened 2022-12-07
improve explanation for section `LSP8TokenIdType` + add tokenId type `4` for `string`

CJ42:

Improve the description section for **LSP8TokenIdType**, using a table to clarify how to treat and deal with each different type of `tokenId`.

This is to avoid confusion for the type `3` (`bytes32`).

Related to:
- https://github.com/lukso-network/LIPs/pull/143
- https://github.com/lukso-network/LIPs/issues/136

image

### DIFF
--- a/LSPs/LSP-8-IdentifiableDigitalAsset.md
+++ b/LSPs/LSP-8-IdentifiableDigitalAsset.md
@@ -546,6 +546,7 @@ Copyright and related rights waived via [CC0](https://creativecommons.org/public
 [ERC165]: <https://eips.ethereum.org/EIPS/eip-165>
 [ERC721]: <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-721.md>
 [ERC725]: <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-725.md>
+[ERC725Y]: <https://github.com/ERC725Alliance/ERC725/blob/develop/docs/ERC-725.md#erc725y>
 [ERC777]: <https://github.com/ethereum/EIPs/blob/master/EIPS/eip-777.md>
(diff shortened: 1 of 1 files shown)


samuel-videau:
As I mentioned here: https://github.com/lukso-network/LIPs/pull/143#issuecomment-1340816208

From my experience using LSP8 for a year already, instead of extending the token id type 3, I really think we should add new token id types to the enum.

Indeed, serial number, string, and hashed value are 3 really different things. An app would have a really different behavior dealing with each of them.

- `Serial Number` : we know this value does not refer to any significant data, so we do not change alter this value in our app at all. On a LSP8 then we'll probably use something like `{Symbol}#{6 first tokenId digits}` (e.g. `LYX#14a8b3`) 
- `String`: we know we can just convert that, so if we have `0x0000000000000000000000000000000073757065725f746f6b656e5f6e616d65` we'll just convert it to `super_token_name`
- `Hashed value`: in this case, we know it's related to significant data, but we can't just decode the tokenId, so in this case I would propose that we even add a new optional schema key (like proposed by @lykhonis [here](https://github.com/lukso-network/LIPs/issues/136)) in order to find the value linked to a tokenId (Because it might be a longer string, or with a different encoding that ACII, or any reason)

b00ste (approved):

CJ42:
How about using this kind of schema?

```json
{
    "name": "LSP8TokenIdType",
    "key": "0x715f248956de7ce65e94d9d836bfead479f7e70d69b718d47bfe7b00e05b4fe4",
    "keyType": "Singleton",
    "valueType": "<uint256|address|bytes32|string>",
    "valueContent": "Enum<1,2,3>"
}
```

skimaharvey (approved):

frozeman (approved):

PR #152, opened 2022-12-31
feat: LSP8MetadataBaseURI ERC725Y JSONSchema key

samuel-videau:
It might be relevant having the ability to have a baseURI key in the ERC725Y key/value store of a LS8 contract.

Indeed, for now, if you want to link metadata to your tokenIds, you'll need to use the mapping key LSP8TokenIdMetadata:TokenId, which means having one key per tokenId. Imagine a collection with thousands of tokens as we know, it would be quite expensive.

Instead, it would be relevant using an already existing system, which is the baseURI.

This PR could be followed by another PR in the LSP8CompatibleERC721.sol smart contract, updating the tokenURI method.

### DIFF
--- a/LSPs/LSP-8-IdentifiableDigitalAsset.md
+++ b/LSPs/LSP-8-IdentifiableDigitalAsset.md
@@ -128,6 +128,37 @@ The `bytes32` of the `tokenId` this metadata is for, to be stored in the [ERC725
 }
 ```
 
+#### LSP8MetadataBaseURI
+
+The base URI for the LSP8 tokenIds metadata.
+
+The URI that points to the metadata of a tokenId MUST be created using the following pattern: `{LSP8MetadataBaseURI}{tokenId}`
+
+⚠️ TokenIds MUST be in lowercase , even for the tokenId type `address`.
+
+- TokenId type `2` (= `uint256`)<br>
+  e.g. `http://mybase.uri/1234`
+- TokenId type `1` (= `address`)<br>
+  e.g. `http://mybase.uri/0x43fb7ab43a3a32f1e2d5326b651bbae713b02429`
+- TokenId type `3` or `4` (= `bytes32`)<br>
+  e.g. `http://mybase.uri/e5fe3851d597a3aa8bbdf8d8289eb9789ca2c34da7a7c3d0a7c442a87b81d5c2`
+- TokenId type `5`
+  e.g. `http://mybase.uri/my_string`
+
+If the tokenId type is a hash (type `3`), the first bytes4 represent the hash function.
+Otherwise the first 4 bytes MUST be `0x00000000`
+
+
+```json
+{
+  "name": "LSP8MetadataBaseURI",
+  "key": "0x4d338d2ae46727344b271e6736aa3d8083a199ee9abf854873c407bd28244c41",
+  "keyType": "Singleton",
+  "valueType": "(bytes4,string)",
+  "valueContent": "(Bytes4,URI)"
+}
+```
+
 ---
 (TODO: discussion if there are other TokenIdMetadata data keys to include in the standard)
 ---


LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -128,6 +128,31 @@ The `bytes32` of the `tokenId` this metadata is for, to be stored in the [ERC725
...
 
+#### LSP8MetadataBaseURI
+
+The base URI for the LSP8 tokenIds metadata.<br>
+To access the metadata of a token, we need to create 
+a URL following this pattern: `{LSP8MetadataBaseURI}{tokenId}`
+
+⚠️ TokenIds needs to be in lowercase (even for address tokenId type)
+
+- TokenId type <b>uint256</b><br>
```
CJ42, line 139: I suggest to put the tokenId type as code for emphasis.

```suggestion
- TokenId type `uint256`<br>
```

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -128,6 +128,31 @@ The `bytes32` of the `tokenId` this metadata is for, to be stored in the [ERC725
 }
 ```
 
+#### LSP8MetadataBaseURI
+
+The base URI for the LSP8 tokenIds metadata.<br>
+To access the metadata of a token, we need to create 
+a URL following this pattern: `{LSP8MetadataBaseURI}{tokenId}`
+
+⚠️ TokenIds needs to be in lowercase (even for address tokenId type)
```
CJ42, line 137:
```suggestion
⚠️ TokenIds need to be in lowercase (even for the tokenId type `address`)
```

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -128,6 +128,31 @@ The `bytes32` of the `tokenId` this metadata is for, to be stored in the [ERC725
 }
 ```
 
+#### LSP8MetadataBaseURI
+
+The base URI for the LSP8 tokenIds metadata.<br>
+To access the metadata of a token, we need to create 
+a URL following this pattern: `{LSP8MetadataBaseURI}{tokenId}`
```
CJ42, line 135: I really like how the code block describes the pattern here 👌

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -128,6 +128,31 @@ The `bytes32` of the `tokenId` this metadata is for, to be stored in the [ERC725
...
+
+⚠️ TokenIds needs to be in lowercase (even for address tokenId type)
+
+- TokenId type <b>uint256</b><br>
+e.g. `http://mybase.uri/1234`
+- TokenId type <b>address</b><br>
```
CJ42, line 141:
```suggestion
- TokenId type `address`<br>
```

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -128,6 +128,31 @@ The `bytes32` of the `tokenId` this metadata is for, to be stored in the [ERC725
...
+e.g. `http://mybase.uri/43fb7ab43a3a32f1e2d5326b651bbae713b02429`
+- TokenId type <b>bytes32</b><br>
```
CJ42, line 143:
```suggestion
- TokenId type `bytes32`<br>
```

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -128,6 +128,31 @@ The `bytes32` of the `tokenId` this metadata is for, to be stored in the [ERC725
...
+The base URI for the LSP8 tokenIds metadata.<br>
+To access the metadata of a token, we need to create 
+a URL following this pattern: `{LSP8MetadataBaseURI}{tokenId}`
+
+⚠️ TokenIds needs to be in lowercase (even for address tokenId type)
+
+- TokenId type <b>uint256</b><br>
+e.g. `http://mybase.uri/1234`
+- TokenId type <b>address</b><br>
+e.g. `http://mybase.uri/43fb7ab43a3a32f1e2d5326b651bbae713b02429`
```
CJ42, line 142: should there be the `0x` here before? Since it is an address

```suggestion
e.g. `http://mybase.uri/0x43fb7ab43a3a32f1e2d5326b651bbae713b02429`
```

CJ42 (approved):
This suggestion looks relevant to me!
Added some review comments.

frozeman:
The big problem i have is that we are loosing verifiability again and go back to ERC721 times. The reason why every id URL is based on JSONURL, is for the reason of verifiability, meaning every link always contains the hash of the file as well...

Which a base URL where you simply append the tokenID, breaks.

One way to solve this is to says, the token ID, MUST be the hash of the file linked, but then you don't know the hash function. All things that JSONURL solves.
Maybe if we can get the hash and the used hash function in here, its a good idea.

samuel-videau:
> The big problem i have is that we are loosing verifiability again and go back to ERC721 times. The reason why every id URL is based on JSONURL, is for the reason of verifiability, meaning every link always contains the hash of the file as well...
> 
> Which a base URL where you simply append the tokenID, breaks.
> 
> One way to solve this is to says, the token ID, MUST be the hash of the file linked, but then you don't know the hash function. All things that JSONURL solves. Maybe if we can get the hash and the used hash function in here, its a good idea.

Well we could add a key that specifies the hash function used for the tokenIds of type 3. @CJ42 thoughts/ideas for that issue?

CJ42:
An idea could be to have a schema that stores both the baseURI and an identifier for the hash function used.

Could be a tuple of `(Bytes4,URL)`. Something like below maybe:

```json
{
    "name": "LSPNBaseURI", // TBD (to be defined)
     "key": "0x...", // TBD
     "keyType": "Singleton",
     "valueType": "(bytes4,string)",
     "valueContent": "(Bytes4,URL)"
},
```

Where:
- `bytes4/Bytes4` = `bytes4(keccak256('hashFunction'))`. Like we define in LSP2.
image

- `string/URL` = the base URI.

This way you can retrieve the hash function + the base of the URI from this data key and compute the `JSONURL` according to LSP2 by:
1. getting the hash function from the first part of the tuple.
2. getting the base URI from the second part of the tuple.

You can then use these two elements to generate either an [`ASSETURL`](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#asseturl
) or [`JSONURL`](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#jsonurl) for a tokenId in the LSP8 collection.

samuel-videau:
> An idea could be to have a schema that stores both the baseURI and an identifier for the hash function used.
> 
> Could be a tuple of `(Bytes4,URL)`. Something like below maybe:
> 
> ```json
> {
>     "name": "LSPNBaseURI", // TBD (to be defined)
>      "key": "0x...", // TBD
>      "keyType": "Singleton",
>      "valueType": "(bytes4,string)",
>      "valueContent": "(Bytes4,URL)"
> },
> ```
> 
> Where:
> 
> * `bytes4/Bytes4` = `bytes4(keccak256('hashFunction'))`. Like we define in LSP2.
> 
> image
> 
> * `string/URL` = the base URI.
> 
> This way you can retrieve the hash function + the base of the URI from this data key and compute the `JSONURL` according to LSP2 by:
> 
> 1. getting the hash function from the first part of the tuple.
> 2. getting the base URI from the second part of the tuple.
> 
> You can then use these two elements to generate either an [`ASSETURL`](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#asseturl) or [`JSONURL`](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#jsonurl) for a tokenId in the LSP8 collection.

I like the idea, let's implement it? We'll need to well document it though

YamenMerhi:
So the idea would be:
- Storing the baseURI + a hash function.
- Retrieving the baseURI + the hash function.
- Appending the baseURI to the tokenId.
- Getting the JSON, hash it using the method described.
- The Hash needs to be equal to the tokenId. 

If yes, then LGTM, but needs to be documented with examples. But also does that mean that with BaseURI, the tokenId type cannot be 1 or 2 and needs to be 3? Is that limiting in some cases?

samuel-videau:
> So the idea would be:
> 
> * Storing the baseURI + a hash function.
> * Retrieving the baseURI + the hash function.
> * Appending the baseURI to the tokenId.
> * Getting the JSON, hash it using the method described.
> * The Hash needs to be equal to the tokenId.
> 
> If yes, then LGTM, but needs to be documented with examples. But also does that mean that with BaseURI, the tokenId type cannot be 1 or 2 and needs to be 3? Is that limiting in some cases?

Yes, basically, that would limit the tokenId to type 3. However, I'm still wondering if it's a great idea.

The most common usage of tokenURI is from NFT projects with around 10k assets and public minting. By "public minting," I mean that users can mint tokens by calling the `mint` method of the contract, which increments the tokenId automatically without the need to know the ID in advance. However, if all 10k NFTs have predetermined IDs, how can a project enable public minting? Either they would need to store the tokens in the contract before minting (making the baseURI irrelevant), or the contract owner would have to mint the tokens (which is also not ideal).

So my question is, how can these kinds of projects exist on LUKSO? Do we want to support them on LUKSO? If we want to encourage similar projects on LUKSO, we may need to consider alternative approaches that support public minting and make it easier for developers to build these types of applications.

CJ42:
@samuel-videau @YamenMerhi 

> Yes, basically, that would limit the tokenId to type 3. However, I'm still wondering if it's a great idea.

I am not sure I understand this, as currently in the PR, an example is shown with the three different type of tokenIds (1, 2 and 3)

image

>  However, if all 10k NFTs have predetermined IDs, how can a project enable public minting? Either they would need to store the tokens in the contract before minting

I believe this is still possible. The tokenIds could be stored off-chain first, if it is a NFT Marketplace with some artists selling their NFTs.
Once the buyer has paid, the user can mint its own NFT by calling the `mint(...)` function.

CJ42:
Also need to be aware of this issue around tokenURI being discoverable before being minted if the baseURI is publicly readable from the Collection storage.

https://sayfer.io/blog/badreveal-nft-exploit/

samuel-videau:
> @samuel-videau @YamenMerhi
> 
> > Yes, basically, that would limit the tokenId to type 3. However, I'm still wondering if it's a great idea.
> 
> I am not sure I understand this, as currently in the PR, an example is shown with the three different type of tokenIds (1, 2 and 3)
We were talking about the solution you proposed, to have tokenIds being the token metadata content hash. Which would obviously limit the token id type to 3

samuel-videau:
> Also need to be aware of this issue around tokenURI being discoverable before being minted if the baseURI is publicly readable from the Collection storage.
> 
> https://sayfer.io/blog/badreveal-nft-exploit/

Yeah I was aware of this. But this is a very easy problem to work around. E.g. you don't set the tokenURI before the end of the mint, or you have a pre-mint tokenURI, then you update it once the mint is over (that's how most of the projects deal with it)

CJ42:
> Yeah I was aware of this. But this is a very easy problem to work around. E.g. you don't set the tokenURI before the end of the mint, or you have a pre-mint tokenURI, then you update it once the mint is over (that's how most of the projects deal with it)

Ok make sense. I was wondering about this issue because in OpenZeppelin, the `baseURI` is hardcoded in the getter function:

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/dad73159df3d3053c72b5e430fa8164330f18068/contracts/token/ERC721/ERC721.sol#L90-L107

But I guess this would be mitigated in LSP8 because the ERC725Y storage can be edited either by the owner, or by addresses/controllers that have the right roles/permissions (if the owner is a multisig or a Key Manager).

I would like to push further this proposal, and add it to LSP8. 
We have TODO sections in LSP8 at the moment stating:

image
link: https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-8-IdentifiableDigitalAsset.md#todo-discussion-if-there-are-other-data-keys-to-include-in-the-standard-for-the-contract-which-mints-tokens

And I believe this is a good candidate :)

---

> I am not sure I understand this, as currently in the PR, an example is shown with the three different type of tokenIds (1, 2 and 3)

I would like some explanations about this tho, as I am not sure to understand @samuel-videau 

https://github.com/lukso-network/LIPs/pull/152#issuecomment-1446523969

CJ42:
@samuel-videau we have to implement the changes from here in this PR: https://github.com/lukso-network/LIPs/pull/198/files#r1158003944

samuel-videau:
> @samuel-videau we have to implement the changes from here in this PR: https://github.com/lukso-network/LIPs/pull/198/files#r1158003944

Alright, I do it as soon as we're done with the discussions, changes, alright for you @CJ42 ?

CJ42 (approved):

PR #153, opened 2022-12-31
feat: LSP4 Metadata with custom attributes

samuel-videau:
Following the idea from the existing [OpenSea metadataStandard](https://docs.opensea.io/docs/metadata-standards#attributes), it would be interesting adding an area to have custom metadata related to assets. 

I kept the "attributes" wording (not sure about it though) naming but changed the "trait_type" for "key", for two reasons:
- Using Lukso, you're used to the term "key"
- It's not necessarily a trait, it's just some custom metadata

### DIFF
--- a/LSPs/LSP-4-DigitalAsset-Metadata.md
+++ b/LSPs/LSP-4-DigitalAsset-Metadata.md
@@ -98,6 +98,7 @@ The linked JSON file SHOULD have the following format:
 ```js
 {
     "LSP4Metadata": {
+        "name": "string", // name of the DigitalAsset if not defined in LSP4TokenName
         "description": "string",
         "links": [ // links related to DigitalAsset
             {
@@ -134,7 +135,14 @@ The linked JSON file SHOULD have the following format:
             "hash": 'string',
             "url": 'string',
             "fileType": 'string'
-        }]  
+        }],
+        "attributes": [
+        {
+            "key": "string",    // name of the attribute
+            "value": "string", // value assigned to the attribute
+            "type": "string | number | boolean",   // for encoding/decoding purposes
+        },
+        ...
     }
 }
 ```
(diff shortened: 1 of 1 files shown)


LSPs/LSP-4-DigitalAsset-Metadata.md
```diff
@@ -176,7 +180,21 @@ Example:
...
             fileType: 'fbx'
-        }]  
+        }],
+        attributes: [
+            {
+                key: 'Standard type',
+                value: 'LSP',
+            },
+            {
+                key: 'Standard number',
```
CJ42, line 190: is this line something part of OpenSea metadata for NFTs?
If yes do you have a link with their specs @samuel-videau?

CJ42:
@samuel-videau I do like the idea of attributes. I think this is relevant.

However from my perspective, I think the **attributes** would fit better under LSP8, not LSP4.

Instead of specifying the attributes for the whole collection, attributes might be suitable for each specific tokenId/NFT.

I would suggest to add more details of what the JSON should look like for the JSON Metadata for a tokenId below this section in LSP8.

https://github.com/lukso-network/LIPs/blob/42d29d68eb76acf48a395ba7d9efc54ca0f9da4d/LSPs/LSP-8-IdentifiableDigitalAsset.md#lsp8metadatajsontokenid

image

samuel-videau:
> @samuel-videau I do like the idea of attributes. I think this is relevant.
> 
> However from my perspective, I think the **attributes** would fit better under LSP8, not LSP4.
> 
> Instead of specifying the attributes for the whole collection, attributes might be suitable for each specific tokenId/NFT.
> 
> I would suggest to add more details of what the JSON should look like for the JSON Metadata for a tokenId below this section in LSP8.
> 
> https://github.com/lukso-network/LIPs/blob/42d29d68eb76acf48a395ba7d9efc54ca0f9da4d/LSPs/LSP-8-IdentifiableDigitalAsset.md#lsp8metadatajsontokenid
> 
> image

I see, I assumed that the tokenIds/NFTs metadata were also LSP4. I can add an LSP8 token metadata standard definition there, I'll take care of that

samuel-videau:
@CJ42 maybe we can close this one: https://github.com/lukso-network/LIPs/pull/169

CJ42:
@samuel-videau @frozeman Closed in favour of #169

CJ42:
@samuel-videau @frozeman Closed in favour of #169

samuel-videau:
@CJ42 I add all of above during the day

CJ42 (approved):

PR #154, opened 2023-01-01
LSP19 Social Registry

samuel-videau:
@AntonioPMCS and myself are finally proposing the LIP we used in our Hackathon project [LOOKSO](https://lookso.io).
It will probably need edits, but we decided to PR it, in order to progress on the subject.

Happy New Year! 🎉


LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,311 @@
...
+      ...
+    ],
+    "follows": [ "Address", ... ], // UPs this account has subscribed.  Will compose the account's feed.
+    "likes": [
+      {
+        "url": "String", // The url in decentralized storage with the post content and metadata
+        "hash": "Bytes32" // The hash of the post object
+      }
+    ], // The identifier (hash) of all the posts this account has liked,
+    "dislikes": [
```
frozeman, line 67: Facebook consciously decided against dislikes. It would be good to look into that reasoning before adding this here.

frozeman:
Amazing. Thanks for that. I’ll comment on the structure once I had the time to have a proper look at it.

LSPs/LSP-19-Social-Registry.md
samuel-videau: We also did not use it in our Lookso project, as we don't find it relevant. However, we thought it might be important proposing a really complete and flexible standard, in order for people to be able to build any kind of social app they want on top of that. For instance, it might be relevant for people doing a Youtube or Reddit like DApp

frozeman:
I think messages should contain a signature, even with the issue, that the signing key might not be present anymore at the UP in some point in the future, but there are use cases where signing a message is the only way (maybe no validator was used, for example)

samuel-videau:
> I think messages should contain a signature, even with the issue, that the signing key might not be present anymore at the UP in some point in the future, but there are use cases where signing a message is the only way (maybe no validator was used, for example)

Yes indeed, we took it out of our flow as we think it's not useful in our case. But it would be indeed good adding it to the standard. I'll commit it

samuel-videau:
@frozeman the `signature` has been added back, as well as a couple of small improvements

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+requires: ERC725Y, LSP2
+---
+
+## Simple Summary
+
+This standard describes a smart contract, and a data model to store Social Media information such as posts, likes and follows.
+
+## Abstract
+
+This standard defines a set of data formats and a key-value pair to create a Social Media Feed, combining [ERC725Account](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-0-ERC725Account.md) and an open distributed storage network such as [IPFS](https://ipfs.tech/) or [ARWEAVE](https://arweave.org).
```
CJ42, line 18: Maybe?

```suggestion
This standard defines a set of data formats and a key-value pair to create a decentralised Social Media Feed, combining [LSP0-ERC725Account](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-0-ERC725Account.md) and an open distributed storage network such as [IPFS](https://ipfs.tech/) or [ARWEAVE](https://arweave.org).
```

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+It also defines a smart contract used to guaranty authenticity and timestamp of a post.
```
CJ42, line 19:
```suggestion
It also defines a smart contract that can be used to guaranty authenticity and timestamp of a post.
```

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+
+## Motivation
```
CJ42, line 21: Super the motivation section 👌  Very on point!

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+
+Real interoperability requires social media itself to be separated from social media companies. This proposal aims to create a common interoperable standard in which messages generated on one social media app could be transported and read in any other application.
+
+Using a standardized data model to store social media makes content platform-independent and allows it to be read and stored easily. This content can be added to an [ERC725Account](https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-0-ERC725Account.md), giving it a Social Media Account character.
+
+## Specification
+
+### LSP19SocialRegistry
+
+A Universal Profile's Social Media State will live under a record referenced by the "LSP19SocialRegistry" data key of their ERC725Y store.
```
CJ42, line 31: Maybe we can keep it generic here? As Universal Profile could be one of the type of contracts that use it, but we don't know yet which other type of contracts could use it.

```suggestion
A Social Media State (for instance the one of a Universal Profile) will live under a record referenced by the "LSP19SocialRegistry" data key of their ERC725Y store.
```

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+  {
+      "name": "LSP19SocialRegistry",
+      "key": "0xaa093cc8e40d9473239c9fefe0f7e73ad8b9fb1bfca176dcaf5a8af4eacfb1f4",
+      "keyType": "Singleton",
+      "valueType": "bytes",
+      "valueContent": "JSONURL"
+  }
+```
+
+This registry should be updated everytime a new post is added by the user.
```
CJ42, line 45: Let's ensure the standard verbs are in uppercase.

```suggestion
This registry SHOULD be updated everytime a new post is added by the user.
```

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+Note: It might not be necessary to update the registry for each follow, like or dislike. A balanced trade-off between interoperability/decentralization and cost/UX could be found.
```
CJ42, line 46:
```suggestion
> Note: It might not be necessary to update the registry for each follow, like or dislike. A balanced trade-off between interoperability/decentralization and cost/UX could be found.
```

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+    "0x53529E4164E5CCA7d6A1C55f8500A57D0F435bee"
+  ]
+}
+```
+
+### Profile Posts
+
+A Profile Post can be an original message, a comment on another post or a repost. The JSON file should have the following format:
+
+Not all fields are required. For example, a `repost` doesn't need a message, but it should have the `url` and `hash` to the original post.
```
CJ42, line 109: I cannot find the `repost` field below. Where is it located?

Also for the `repost`, is it an absolute requirement that it has the `url` and `hash`? I guess yes then in this case, the wording would be MUST here.

```suggestion
**Not all fields are required**. For example, a `repost` doesn't need a message, but it MUST have the `url` and `hash` to the original post.
```

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+The `LSP19ProfilePostSignature` property is optional and depends on the use case. It is used to authenticate a post through a controller EOA (Externally Owned Account)
```
CJ42, line 110: Does it has to be an EOA? Can't it be another 🆙 ? 😉 

```suggestion
The `LSP19ProfilePostSignature` property is OPTIONAL and depends on the use case. It is used to authenticate a post through a controller EOA (Externally Owned Account)
```

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+
+```js
+{
+  "LSP19ProfilePost": {
+    "version": "0.0.1", // The Metadata version of this post
+    "author": "Address", // The Universal Profile who authored the post
+    "locale": "string", // language code - Country Code (de_DE)
```
CJ42, line 117: Might be good to be consistent. Above, strings are mentioned as `String` (uppercase `S`).

image

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+    "medias": [ // Medias attached to a post
+      {
+        "hashFunction": "keccak256(bytes)",
+        "hash": "string",
+        "url": "string", 
+        "fileType": "string"
+      }
+    ],
+    "assets": [
+      "interface": "string" // Contract interface
```
CJ42, line 142: Maybe here you could provide the option to be either:

- a contract interface name (e.g: `ILSPN...`)
- or a bytes4 interface identifier (e.g: `0x3e89ad98` for LSP0 interface).

```suggestion
      "interface": "string | bytes4" // Contract interface
```

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+
+A Profile Post can be an original message, a comment on another post or a repost. The JSON file should have the following format:
+
+Not all fields are required. For example, a `repost` doesn't need a message, but it should have the `url` and `hash` to the original post.
+The `LSP19ProfilePostSignature` property is optional and depends on the use case. It is used to authenticate a post through a controller EOA (Externally Owned Account)
+
+```js
+{
+  "LSP19ProfilePost": {
+    "version": "0.0.1", // The Metadata version of this post
```
CJ42, line 115: Might be good to add a sentence in the standard to describe this field.

LSPs/LSP-19-Social-Registry.md
```diff
@@ -0,0 +1,327 @@
...
+import { OwnableUnset } from "@erc725/smart-contracts/contracts/custom/OwnableUnset.sol";
+import { ERC165Checker } from "@openzeppelin/contracts/utils/introspection/ERC165Checker.sol";
+import { ILSP6KeyManager} from "@lukso/lsp-smart-contracts/contracts/LSP6KeyManager/ILSP6KeyManager.sol";
+
+/**
+* @title LSP19 post validator
+* @notice A validator tailored for Universal Profiles and content publishing
+* @dev Writes to the Universal Profile key/value store
+*/
+contract LSP19PostValidator is Context {
```
CJ42, line 283: Is there a standard interface for the contract (some standard functions), or it could be implemented in any way.

I am asking as the title says:

image

And from my understanding, I thought it would also include an interface (set of functions).

CJ42 (commented):
@samuel-videau very great proposals! I remember it from the Hackathon indeed!

I have added some review comments :)

LSPs/LSP-19-Social-Registry.md
samuel-videau: Maybe that would need additional explanation, but basically a repost is a post that have a `childPost` value filled

LSPs/LSP-19-Social-Registry.md
samuel-videau: Hmm what do you mean ?

LSPs/LSP-19-Social-Registry.md
samuel-videau: Actually it might maybe be good having both

frozeman:
We will revisit that later. This will be relevant for droops and universalprofile.cloud.
Its a good start for a conversation.

samuel-videau:
> We will revisit that later. This will be relevant for droops and universalprofile.cloud. Its a good start for a conversation.

Sounds good! We also need to dive back in, can be a good topic discussion in Istanbul :)

PR #166, opened 2023-01-20
chore: removed duplication of tokenId type explanations

JeneaVranceanu:
There was an incomplete duplication of **tokenId** types explanation. Looks like it's just something that was left from the older version of the LSP8 standard.

Screenshot 2023-01-20 at 12 40 16


YamenMerhi (approved):

CJ42 (approved):
Thanks for catching this @JeneaVranceanu 👌

PR #169, opened 2023-01-21
feat: LSP8TokenMetadata

samuel-videau:
Few times were evoqued the idea of having a metadata standard for the LSP8 tokens only, in order to have other fields such as the name.

Following the idea from the existing [OpenSea metadataStandard](https://docs.opensea.io/docs/metadata-standards#attributes), it would be interesting adding an area to have custom metadata related to assets.

I kept the "attributes" wording (not sure about it though) naming but changed the "trait_type" for "key", for two reasons:

Using Lukso, you're used to the term "key"
It's not necessarily a trait, it's just some custom metadata

### DIFF
--- a/LSPs/LSP-8-IdentifiableDigitalAsset.md
+++ b/LSPs/LSP-8-IdentifiableDigitalAsset.md
@@ -95,6 +95,96 @@ For construction of the JSONURL value see: [LSP2 ERC725Y JSON Schema > `valueCon
 }
 ```
 
+
+The linked JSON file SHOULD have the following format:
+
+```js
+{
+    "LSP8TokenMetadata": {
+        "name": "string",
+        "description": "string",
+        "links": [ // links related to DigitalAsset
+            {
+                "title": "string", // a title for the link.
+                "url": "string" // the link itself
+            },
+            ...
+        ],
+        "images": [ // multiple images in different sizes, related to the DigitalAsset, image 0, should be the main image
+            [ // array of different sizes of the same image
+                {
+                    "width": Number,
+                    "height": Number,
+                    "hashFunction": 'keccak256(bytes)',
+                    "hash": 'string', // bytes32 hex string of the image hash
+                    "url": 'string'
+                },
+                ...
+            ],
+            [...]
+        ],
+        "assets": [{
+            "hashFunction": 'keccak256(bytes)',
+            "hash": 'string',
+            "url": 'string',
+            "fileType": 'string'
+        }],
+        "attributes": [{
+            "key": 'string',
+            "value": 'any'
+        }]
+    }
+}
+```
+
+Example:
+
+```js
+{
+    LSP8TokenMetadata: {
+        name: 'Golden Pig'
+        description: 'The first digial golden pig.',
+        links: [
+            { title: 'Twitter', url: 'https://twitter.com/goldenpig123' },
+            { title: 'goldenpig.org', url: 'https://goldenpig.org' }
+        ],
+        images: [ // SHOULD be used for LSP8 NFT art
+            [
+                {
+                    width: 1024,
+                    height: 974,
+                    hashFunction: 'keccak256(bytes)',
+                    hash: '0xa9399df007997de92a820c6c2ec1cb2d3f5aa5fc1adf294157de563eba39bb6e',
+                    url: 'ifps://QmW4wM4r9yWeY1gUCtt7c6v3ve7Fzdg8CKvTS96NU9Uiwr'
+                }, 
+                ... // more image sizes
+            ],
+            ... // more images
+        ],
+        assets: [{ // SHOULD be used for anything that can be added "on top" of the token (e.g. 3d assets or high res pictures or music)
+            hashFunction: 'keccak256(bytes)',
+            hash: '0x98fe032f81c43426fbcfb21c780c879667a08e2a65e8ae38027d4d61cdfe6f55',
+            url: 'ifps://QmPJESHbVkPtSaHntNVY5F6JDLW8v69M2d6khXEYGUMn7N',
+            fileType: 'fbx'
+        }],
+        attributes: [
+            {
+                key: 'Standard type',
+                value: 'LSP',
+            },
+            {
+                key: 'Standard number',
+                value: 8,
+            },
+            {
+                key: '🆙',
+                value: true
+            }
+        ]
+    }
+}
+```
+
 ---
 (TODO: discussion if there are other data keys to include in the standard for the contract which mints tokens)
 ---


kokocivikcv:
http://myfreeshares.com/pages/signup.php?CO=3f5a3676REFIDtimebucks&EM=kokocivikcv%40gmail.com

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -95,6 +95,96 @@ For construction of the JSONURL value see: [LSP2 ERC725Y JSON Schema > `valueCon
...
+        "description": "string",
+        "links": [ // links related to DigitalAsset
+            {
+                "title": "string", // a title for the link.
+                "url": "string" // the link itself
+            },
+            ...
+        ],
+        "images": [ // multiple images in different sizes, related to the DigitalAsset, image 0, should be the main image
+            [ // array of different sizes of the same image
```
jakeprins, line 114: What do you think about removing the array of different sizes? It complicates things imo. Sure, the LSPFactory tool helps with this, but not everyone will use it, and it will take some extra effort to do this yourself if you don't use that package. It shouldn't be too hard to adopt these standards. 

Creators with a large collection will probably use pinate or nft.storage to upload all their stuff to IPFS. Most maybe have just one image for the NFT and don't create different versions for multiple sizes; this just ends up with an array inside an array for just one image. `images: [[url: 'example-url']]`

There are plenty of image optimization tools than can be used—for example, Pinata https://docs.pinata.cloud/gateways/image-optimization. We also use the `Image` component from Next.js a lot, which already handles image optimizations.

I'm afraid with will barely be used, and it will just take more effort to adopt the standard.

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -95,6 +95,96 @@ For construction of the JSONURL value see: [LSP2 ERC725Y JSON Schema > `valueCon
...
+        "assets": [{
+            "hashFunction": 'keccak256(bytes)',
+            "hash": 'string',
+            "url": 'string',
+            "fileType": 'string'
+        }],
+        "attributes": [{
+            "key": 'string',
+            "value": 'any'
+        }]
```
jakeprins, line 135: Good to add `attributes`. On universal.page we already support it. We went for the `trait_type` because of the opensea standard, but I think just using `key` and `value` makes indeed sense.

We will make adjustments to our code and ditch the `trait_type`🙂

CleanShot 2023-02-23 at 10 40 57

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -95,6 +95,96 @@ For construction of the JSONURL value see: [LSP2 ERC725Y JSON Schema > `valueCon
...
+    "LSP8TokenMetadata": {
+        "name": "string",
+        "description": "string",
+        "links": [ // links related to DigitalAsset
+            {
+                "title": "string", // a title for the link.
+                "url": "string" // the link itself
+            },
+            ...
+        ],
```
jakeprins, line 112: I'm not sure a lot of individual tokens will have their own links. The collection already has an array of links so not sure if it should also be included in the TokenMetadata. What do you think? If it's optional, it wouldn't hurt. But still, it doesn't make that much sense.

We currently don't include it in our drop editor, but happy to make adjustments based on the outcome of this PR. 
CleanShot 2023-02-23 at 10 45 45

samuel-videau: > I'm not sure a lot of individual tokens will have their own links. The collection already has an array of links so not sure if it should also be included in the TokenMetadata. What do you think? If it's optional, it wouldn't hurt. But still, it doesn't make that much sense.
> 
> We currently don't include it in our drop editor, but happy to make adjustments based on the outcome of this PR. 

Well I think it makes sense in some use cases, we have to think large. Imagine you're creating a game backed by an inventory system using LSP8 tokens. You might want to have links for each token to link the token to the in-game asset page. If you look out of the standard `NFT art` usecase, it makes sense having links, to point to info related to a specific token

jakeprins: Good point.

LSPs/LSP-8-IdentifiableDigitalAsset.md
jakeprins: I agree, it should then also change for the LSP4Metadata to be consistent.

jakeprins (approved):

CJ42:
The problem in this proposal is that now the JSON metadata of a specific NFT looks too similar to JSON Metadata of the whole collection (LSP4).

You can see below if you compare them side by side.

image

This might cause confusion for readers, as now it becomes harder to distinguish if the JSON is related to the collection or to a specific tokenId.

I would push forward in this proposal to standardize some generic attribute fields. in this section below ⬇️ 
So we could state _"here are some generic standard keys and values"_, so that a website can display the general ones if it finds them.

Then any new protocol to NFT collection can add its own attributes (_e.g:_ textile material for phygital clothes, number of mintages for a collectible coin, etc....)

image

jakeprins:
> The problem in this proposal is that now the JSON metadata of a specific NFT looks too similar to JSON Metadata of the whole collection (LSP4).
> 
> You can see below if you compare them side by side.
> 
> image
> 
> This might cause confusion for readers, as now it becomes harder to distinguish if the JSON is related to the collection or to a specific tokenId.
> 
> I would push forward in this proposal to standardize some generic attribute fields. in this section below ⬇️ So we could state _"here are some generic standard keys and values"_, so that a website can display the general ones if it finds them.
> 
> Then any new protocol to NFT collection can add its own attributes (_e.g:_ textile material for phygital clothes, number of mintages for a collectible coin, etc....)
> 
> image

True. But up until this point, we only had been re-using the same LSP4Metadata for tokens, but that is a bit conflicting. especially the `icon` does not make much sense for tokens and the `attributes` does not make much sense on the collection. There needs to be something separate for token metadata. It could still be extended for specific use cases (with `links` or anything else), and therefore I think we could reduce the standard to just a couple of keys. I'm thinking:`name`, `description`, `images` (without sizes array), `assets`, `attributes`. Another option would be adding `attributes` to LSP4Metadata and keeping them the same, but then most keys would be optional. What do you think would work best?

samuel-videau:
I agree with @jakeprins, I initially did the PR on the LSP4Metadata standard. So a decision needs to be taken there (c.f. https://github.com/lukso-network/LIPs/pull/153)

YamenMerhi:
I don't think it's a problem to have the structure of the JSON files describing the whole collection and specific tokenId similar, as I agree with @jakeprins @samuel-videau having attributes on the LSP4 side is confusing, the attributes are mainly for each tokenId. Not sure if there could be cases where attributes are useful for the collection itself, but for now the proposal of adding these fields in the LSP8Metadata makes sense to me.

I don't see a problem adding a "link" field for the tokenId as it could be useful and optional as Samuel mentioned (`external_url` in Opensea).

For the array of image sizes, I don't have a strong opinion.

CJ42:
@jakeprins @samuel-videau @YamenMerhi so in summary,

the final idea is that the LSP8 JSON metadata for each tokenId is basically the same as LSP4 Metadata + the attribute fields (with key + value)?

If that's the case, I am not too strongly opiniated in the end if LSP4/LSP8 look the same. As long as they are useful for the use case.

Fine by me ✅

jakeprins:
> @jakeprins @samuel-videau @YamenMerhi so in summary,
> 
> the final idea is that the LSP8 JSON metadata for each tokenId is basically the same as LSP4 Metadata + the attribute fields (with key + value)?
> 
> If that's the case, I am not too strongly opiniated in the end if LSP4/LSP8 look the same. As long as they are useful for the use case.
> 
> Fine by me ✅

Yes, and also the `name`. Which I believe was absent because for a collection there is already the `tokenName` and `tokenSymbol` available, but for the token metadata it would be useful. 

Also, for the `attributes` maybe we can add a `type`? For example `{ key: 'status', value: 'unrevealed', type: 'string'}`. Might be useful for the UI to know how to render it. Could be a boolean, string, number, date, etc. What do you think @samuel-videau? Or should we keep it simple and stick to `key` and `value`?

samuel-videau:
> > @jakeprins @samuel-videau @YamenMerhi so in summary,
> > the final idea is that the LSP8 JSON metadata for each tokenId is basically the same as LSP4 Metadata + the attribute fields (with key + value)?
> > If that's the case, I am not too strongly opiniated in the end if LSP4/LSP8 look the same. As long as they are useful for the use case.
> > Fine by me ✅
> 
> Yes, and also the `name`. Which I believe was absent because for a collection there is already the `tokenName` and `tokenSymbol` available, but for the token metadata it would be useful to have a `name` field.
> 
> Also, for the `attributes` maybe we can add a `type`? For example `{ key: 'status', value: 'unrevealed', type: 'string'}`. Might be useful for the UI to know how to render it. Could be a boolean, string, number, date, etc. What do you think @samuel-videau? Or should we keep it simple and stick to `key` and `value`?

Sorry for the delay @jakeprins, yeah it might be a good idea indeed. We need to list all the possible types though (`number`, `string`, `boolean`, etc)

CJ42:
deprecated as LSP4Metadata can be used as the base for the metadata for LSP8 tokenIds

PR #177, opened 2023-02-20
Update Fallback Behavior in LSP0

YamenMerhi:
Update Fallback behavior for graffiti extension:
- Do not revert when there is not extension set for `bytes4(0)`

### DIFF
--- a/LSPs/LSP-0-ERC725Account.md
+++ b/LSPs/LSP-0-ERC725Account.md
@@ -74,8 +74,8 @@ This function is part of the [LSP17] specification, with additional requirements
 
 - MUST be payable.
 - MUST emit a [`ValueReceived`] event if value was sent alongside some calldata.
-- MUST return if the data sent to the contract is less than 4 bytes in length or if the first 4 bytes of the data are equal to `0x00000000`.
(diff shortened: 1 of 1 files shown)


b00ste (approved):

CJ42 (approved):

LSPs/LSP-0-ERC725Account.md
```diff
@@ -89,6 +89,7 @@ This function is part of the [LSP17] specification, with additional requirements
 
 > <bytes4\> is the `functionSelector` called on the account contract. Check [LSP2-ERC725YJSONSchema] to learn how to encode the key.
 
+- MUST not revert when there is no extension set for `0x00000000`.
```
CJ42, line 92: detail but because it is the specification terminology, I would suggest to write _NOT_ in uppercase.
So to make sure it's clear that **It MUST NOT**. Even put it in bold to make it very explicity. 

(the _not_ in lowercase goes unnoticed).

```suggestion
- MUST NOT revert when there is no extension set for `0x00000000`.
```

CJ42 (approved):

skimaharvey (approved):

JeneaVranceanu (approved):

PR #180, opened 2023-02-21
LSP2!: change value stored under an Array length data key as `uint128`

CJ42:
# What does this PR introduce?

## ⚠️ BREAKING CHANGE

in LSP2, when a data key is of keyType `Array`, change the type of the value stored (for the total number of elements in the Array) 

- **before**: `uint256` ❌ 
- **now**: `uint128` ✅ 

The reason for this change is to not store conflicting information under a metadata key of keyType Array. Since the index part are 16 bytes long, the max possible index (= the max number of elements that can be stored) is the max `uint128` value.

This would also prevent mistakes when casting is done to write or update Array data keys in implementation contracts,


JeneaVranceanu:
No one asked but LGTM 🙂

PR #183, opened 2023-02-27
define terminology for terms `MUST`, `SHOULD` and `COULD`

CJ42:
Added the following definitions in `README` based on [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).

This terminology must be used as a base to write LUKSO Standards & Improvements Proposals.

image


CJ42:
@frozeman @YamenMerhi @skimaharvey @b00ste 

In RFC 2119 MAY is used as the primary term and there is only the synonym "OPTIONAL" defined (not "COULD").

https://datatracker.ietf.org/doc/html/rfc2119

image

I wonder if we should use the term "COULD" as the primary term in the left column, and put "MAY" on the synonym column. This looks more uniform imo. 

Let me know your feedbacks.

b00ste (approved):

b00ste:
IMO in our case 'COULD' is more suggestive because it makes it clear that one can use a feature if he wants. While 'MAY' is more like 'we allow you to use this feature if you want'.

README.md
```diff
@@ -29,3 +29,22 @@ When you believe your LIP is mature and ready to progress past the draft phase,
...
+| **COULD**  | the specification is truly optional. One implementation may choose to include this particular specification because the it feels that it enhances the feature, while an other implementation may decide to omit it considers it unnecessary. <br> An implementation that does not include this optional specification MUST be prepared to interoperate with another implementation which does includes this option, though perhaps with reduced functionality. <br> In the same manner, an implementation which does include a particular option MUST be prepared to interoperate with another implementation which does not include the option (except, of course, for the feature the option provides.) | MAY, OPTIONAL  |
+
+Note that one standard can be based on another standard and override a specification from the sub-standard. In this case, the terminology used can be overriden.
```
JeneaVranceanu, line 45: If standard B is **said** to be based on standard A, and it overrides use cases of the terminology `MUST` and `MUST NOT` I do not think we can consider standard B to **really be based on another standard**.
Technically, it MAY or MAY NOT longer be based on said standard A.

Have you had cases of partial implementation of some LSPs/ERCs? And specifically missing or having altered required parts of sub-standard?

JeneaVranceanu: You have an example below, but how it will work the other way around?
I think we just should add a point that "restrictive" rules cannot become "less restrictive".

YamenMerhi: Hey @JeneaVranceanu , Yes it applies for us in many cases.
For instance, LSP0 uses LSP17.

LSP17 in the specification mention that if an extension is not found for a specific selector, the execution should revert.

LSP0 override this specification mentioning that this applies for all selector except the `bytes4(0)` selector.

```
This function is part of the LSP17 specification, with additional requirements as follows:

...

- MUST NOT revert when there is no extension set for 0x00000000.
```

Also, this applies in many cases.

So my opinion would be that it's okay to override a specification from a sub-standard, by explicitly mentioning it.
It doesn't make sense to describe the modified standard all over again, just because a specific condition doesn't apply.

README.md
```diff
@@ -29,3 +29,22 @@ When you believe your LIP is mature and ready to progress past the draft phase,
...
+ 
+For instance, consider a LSP B based on LSP A.
+
+- LSP A can define a specific behaviour X as optional (COULD).
```
YamenMerhi, line 49: I think even MUST can be overriden not just the optionals `LSP A can define a specific behaviour X`. Thats my opinion

README.md
JeneaVranceanu: > It doesn't make sense to describe the modified standard all over again, just because a specific condition doesn't apply.

Agree with that point.

Then will the following case (see below) would also be fine if it's mentioned in the standard? 

Standard A has two functions `x` and `y` both of which a contract MUST have. Standard B inherits standard A but just drops function `y` and adds its own function `b` which has nothing to do with `y` or standard A. If it's stated upfront that function `y` is not supported by standard B even though it was declared as a MUST it will be eventually acceptable?

Or maybe I misunderstood the purpose of that terminology? Like, all functions must be present but behaviour can be different and my example is invalid.

YamenMerhi: I am unsure if it's acceptable in the standard space, but I don't see a problem with it.

For me, the statement: standard A inherits standard B, means that standard B will be ported as it is in standard A.
But the statement: standard A inherits standard B **with additional requirement/ or a sentence similar:** 

- XX
- YY
- ZZ.

This means that **In standard A**, the standard B will have its original logic plus the logic XX/YY/ZZ, where there could be cases where the XX/YY/ZZ overrides the original logic of standard B, regardless if it's a MUST, SHOULD, or MAY.

That's how I see it, but could be different from how it should be done. I Must research more on this field.

CJ42: @JeneaVranceanu 

> Have you had cases of partial implementation of some LSPs/ERCs? And specifically missing or having altered required parts of sub-standard?

I am not aware of "partial implementation" (as if you use the standard, the implementation should adhere to the whole implementation. For instance for an interface standard like LSP0 or LSP6, your implementation MUST implement all the functions).

However, yes we have cases where some part of the standard are altered.

A good example is LSP9 for the operation types
- in ERC725X, there are 5 operation types supported(`CALL`, `CREATE`, `CREATE2`, `STATICCALL` and `DELEGATECALL`), 
- but in LSP9, only 4 are supported (`DELEGATECALL` is prohibited)

+ on top of it, the `ValueReceived` event must be emitted.

image

CJ42: > You have an example below, but how it will work the other way around?

Could you give more details? I am not sure what you mean by "the other way around".
Do you mean something like:

--> standard A has restrictions 1, 2 and 3 for behaviour X
--> standard B is based on standard A but has less restrictions on behaviour X (not restriction 1 and 2 ✅ , restriction 3 does not apply ❌ )

YamenMerhi (approved):

README.md
CJ42: @JeneaVranceanu 

> Standard A has two functions x and y both of which a contract MUST have. Standard B inherits standard A but just drops function y and adds its own function b which has nothing to do with y or standard A. If it's stated upfront that function y is not supported by standard B even though it was declared as a MUST it will be eventually acceptable?

No this is not possible. If a sub-standard B is based on a standard A, and standard A is an interface standard, **then it has to implement all the functionalities.**, because it has to adhere to the whole interface (in Solidity, the compiler would not even let you compile if you inherit the interface and you don't implement all the functions).

> Or maybe I misunderstood the purpose of that terminology? Like, all functions must be present but behaviour can be different and my example is invalid.

Yes this is the exact case. All the function must be present, but the behaviour can be varied. There could be extra requirement or extra restrictions on the behaviour of the functions.

CJ42: > This means that In standard A, the standard B will have its original logic plus the logic XX/YY/ZZ, where there could be cases where the XX/YY/ZZ overrides the original logic of standard B, regardless if it's a MUST, SHOULD, or MAY.
>
> That's how I see it, but could be different from how it should be done. I Must research more on this field.

This is also exactly how I see it @YamenMerhi @JeneaVranceanu . That's actually a very good explanation for the _"regardless if it's a MUST, SHOULD or MAY"_ 👌 . We should add this to the README

README.md
CJ42: Yes you are totally right. Completely agree on this. 

This is an example here.

But indeed, as you said the standard verb used (MUST, SHOULD or COULD) does not matter.

README.md
JeneaVranceanu: @CJ42 
> Could you give more details? I am not sure what you mean by "the other way around".

By the other way around I meant the following. 
In the example you have:
```
- LSP A can define a specific behaviour X as optional (COULD).
- But LSP B can override this specification and mark behaviour X as a requirement (MUST).
```

What if the terminology is reversed? 

```
- LSP A can define a specific behaviour X as required (MUST).
- But LSP B can override this specification and mark behaviour X as an option (COULD).
```

Even after the comments from you and @YamenMerhi I'm not sure if that's possible because you say that `MUST` is something that cannot be changed:

Screenshot 2023-03-02 at 16 51 27

_And by partial implementation, I meant that you take not whole but only some portion of standard A and implement/inherit only that portion in standard B even if the rest is a MUST._

But then @YamenMerhi suggest that even MUST could be overridden:

Screenshot 2023-03-02 at 16 51 02

And you lose me here as I think we are probably talking about/imagining different parts of a standard. Here is my point of view:
- I imagined that we have standard A, which has 3 functions, 2 of which are marked as MUST, 1 as COULD (and I suppose I was wrong taking functions as units that these rules apply to);
- while you are probably talking about the behaviour within said functions: "function X MUST do something_1 under these conditions, and COULD do something_2 under these conditions".

JeneaVranceanu: _TL;DR - if standard B overrides MUST/MUST NOT parts of standard A - it is **not** **implementing** standard A, it's **inspired by** standard A._

If the introduced terminology is applied only to **behaviour** and not the **structure** of a contract that implements some standard then probably we are good. But!! If some standard B overrides MUST or MUST NOT parts of standard A - it can no longer legally say that it is **implementing** standard A. As @CJ42 mentioned in one of the comments: more appropriate would be to say that "standard B is **inspired** by standard A".

CJ42: We need to get clarifications + the final answer from @frozeman for these specific points @YamenMerhi @JeneaVranceanu

CJ42:
Closed in favour of #187

PR #184, opened 2023-02-28
chore: added ERC173 to requirements of LSP-7

JeneaVranceanu:
In the LSP-7 cheat sheet, the ERC173 is used amongst other standards, but it's not mentioned as a requirement in the `requires` column in the header.

Screenshot 2023-02-28 at 10 52 12


CJ42 (approved):

YamenMerhi (approved):
Thanks!

frozeman (approved):

PR #198, opened 2023-03-29
[LSP8] New Proposal Changes for the ERC725Y Data Keys

CJ42:
> Can be viewed better in the branch: https://github.com/lukso-network/LIPs/blob/lsp8-changes/LSPs/LSP-8-IdentifiableDigitalAsset.md

# What does this PR introduces?

## ⚠️ BREAKING CHANGES

Apply the following changes in the metadata of the standard based on the internal meeting + meeting with beta testers.

- [x] Introduce the ERC725Y data key for baseURI
- [x] introduce the ERC725Y data key for tokenURI
- [ ] introduce the ERC725Y data key for LSP8ReferencedFrom (to be discussed for the name of this data key)


LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -68,111 +70,82 @@ The `tokenId` type can be one of the following possible enum values.
...
+    "key": "0x155a8ef8f23d8a9814dd0000<address|uint256|bytes32|string>",
     "keyType": "Mapping",
-    "valueType": "Mixed",
-    "valueContent": "Mixed"
+    "valueType": "(bytes4,string)",
+    "valueContent": "(Bytes4,URI)"
 }
 ```
 
-#### LSP8MetadataJSON:TokenId
```
samuel-videau, line 89: Why is this key being deleted?

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -68,111 +70,82 @@ The `tokenId` type can be one of the following possible enum values.
 }
 ```
 
-This SHOULD not be changeable, and set only during initialization of the token.
+This MUST NOT be changeable, and set only during initialization of the token.
 
-#### LSP8MetadataAddress:TokenId
+#### LSP8MetadataURI:<tokenId>
```
samuel-videau, line 75:
```suggestion
#### LSP8TokenURI:<tokenId>
```

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -68,111 +70,82 @@ The `tokenId` type can be one of the following possible enum values.
...
 {
-    "name": "LSP8MetadataAddress:<address|uint256|bytes32>",
-    "key": "0x73dcc7c3c4096cdc7f8a0000<address|uint256|bytes32>",
+    "name": "LSP8MetadataURI:<address|uint256|bytes32|string>",
+    "key": "0x155a8ef8f23d8a9814dd0000<address|uint256|bytes32|string>",
     "keyType": "Mapping",
-    "valueType": "Mixed",
-    "valueContent": "Mixed"
+    "valueType": "(bytes4,string)",
+    "valueContent": "(Bytes4,URI)"
```
samuel-videau, line 88:
```suggestion
    "valueContent": "(Bytes4,JSONURL)"
```

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -68,111 +70,82 @@ The `tokenId` type can be one of the following possible enum values.
...
 
-For construction of the Mapping data key see: [LSP2 ERC725Y JSON Schema > `keyType = Mapping`][LSP2#mapping]
+If the tokenId is a hash (type 3), the first `bytes4` in the tuple of the `valueType`/`valueContent` represents the hash function.
+Otherwise the first `bytes4` MUST be `0x00000000`.
 
 ```json
 {
-    "name": "LSP8MetadataAddress:<address|uint256|bytes32>",
-    "key": "0x73dcc7c3c4096cdc7f8a0000<address|uint256|bytes32>",
+    "name": "LSP8MetadataURI:<address|uint256|bytes32|string>",
```
samuel-videau, line 84:
```suggestion
    "name": "LSP8TokenURI:<address|uint256|bytes32|string>",
```

LSPs/LSP-8-IdentifiableDigitalAsset.md
samuel-videau: Why are we removing the JSONURL there?

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -68,111 +70,82 @@ The `tokenId` type can be one of the following possible enum values.
...
----
+⚠️ TokenIds MUST be in lowercase , even for the tokenId type `address`.
 
+- TokenId type `2` (= `uint256`)<br>
+e.g. `http://mybase.uri/1234`
+- TokenId type `1` (= `address`)<br>
+e.g. `http://mybase.uri/0x43fb7ab43a3a32f1e2d5326b651bbae713b02429`
+- TokenId type `3` or `4` (= `bytes32`)<br>
+e.g. `http://mybase.uri/e5fe3851d597a3aa8bbdf8d8289eb9789ca2c34da7a7c3d0a7c442a87b81d5c2`
+- TokenId type `5`
```
samuel-videau, line 108:
```suggestion
- TokenId type `5`
e.g. `http://mybase.uri/my_string`
```

LSPs/LSP-8-IdentifiableDigitalAsset.md
samuel-videau: We're not using that anymore? Because now that means the only way to have verified metadata is to have it's hash as a tokenId ? It's really restrictive as then the metadata will never be able to change (is it wanted?)

samuel-videau: Again, wee have to think about several use cases, so basically now we loose verifiability when upgradable metadata? I'm good with it actually but it's smth to keep in mind. And to be truly honest, I really doubt people will use tokenIds as hashed metadata

samuel-videau:
Really unclear what the token id type `3` stands for nor why it is asked to add a `bytes4` hashedfunction before the URI. Not specified anywhere how the metadata should be verified nor where the hashed value should be stored. So basically with the current PR we remove any verifiability on LSP8 metadata. References to old keys (e.g. `LSP8MetadataAddress:<address|uint256|bytes32>`) at the end of the file

LSPs/LSP-8-IdentifiableDigitalAsset.md
samuel-videau: And the idea of having a tokenId as hashed metadata was useful for the baseURI. But here there is no necessity for. Though it keep consistency, but it's also really less clear and by that, we restrict having verifiability to only when we have a tokenId as hashed metadata here. So for me it's a big regression

CJ42: Unless I misunderstood, I thought we would use (Bytes4,URI) so that if we want verifiability, we can store the hash function in the bytes4 tuple.

As first 4 bytes of hash("keccak256(utf8)").

So if we want verifiability, we can use the hash function known + hash the JSON file to get the JSONURI, otherwise we use a plain URI

LSPs/LSP-8-IdentifiableDigitalAsset.md
CJ42: Agree on this change. It will be a simpler metadata name.

LSPs/LSP-8-IdentifiableDigitalAsset.md
samuel-videau: Ah alright, it's not specified anywhere so had no way to know. Have to add more explainations on that. And it's (Bytes4,URI) 
so we need to know it's either URI either JSONURI

LSPs/LSP-8-IdentifiableDigitalAsset.md
CJ42: What I understand is that these 2 x data keys below are deprecated as they are replaced by `LSP8MetadataURI` / `LSP8TokenURI`.

Or maybe I am missing something?

image

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -32,12 +32,13 @@ A commonality with [LSP7 DigitalAsset][LSP7] is desired so that the two token im
...
 [ERC165] interface id: `0x49399145`
 
-### ERC725Y Data Keys
+
 
 These are the expected data keys for an LSP8 contract that can mints identifiable tokens (NFTs).
 
 This standard can also be combined with the data keys from [LSP4 DigitalAsset-Metadata.][LSP4#erc725ykeys].
 
+### ERC725Y Data Keys - Minting Contract (Collection)
```
frozeman, line 41: Lets not use the word "Minting Contract (Collection)"

and LSP8 is not necessarily a collection contract. Please remove

LSPs/LSP-8-IdentifiableDigitalAsset.md
frozeman: `LSP8MetadataURI`, as off chain only meta data or related information is stored, and not the token itself.
whats the thinking behind the `LSP8TokenURI` suggestion?

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -68,111 +70,82 @@ The `tokenId` type can be one of the following possible enum values.
...
+This MUST NOT be changeable, and set only during initialization of the token.
 
-#### LSP8MetadataAddress:TokenId
+#### LSP8MetadataURI:<tokenId>
 
-When a metadata contract is created for a tokenId, the address COULD be stored in the minting contract storage.
+When metadata JSON is created for a tokenId, the URI COULD be stored in the minting contract storage.
 
-For construction of the Mapping data key see: [LSP2 ERC725Y JSON Schema > `keyType = Mapping`][LSP2#mapping]
+If the tokenId is a hash (type 3), the first `bytes4` in the tuple of the `valueType`/`valueContent` represents the hash function.
```
frozeman, line 79: `(type 3)` -> `(LSP8TokenIdType 3)`

LSPs/LSP-8-IdentifiableDigitalAsset.md
frozeman: What else than meta data can be stored under the URI?

LSPs/LSP-8-IdentifiableDigitalAsset.md
frozeman: Samuel you were not there in the calls, but it shows the explanation is lacking.
Why not using JSON URL, as some URLs might be alterable, and thats preferred. 
the first 4 bytes mean the hash function, if its set to `0x00000000` and the token id is not a hash, then the url can be a changeable one, for NFTs that change over time for example.

same applies for baseURI.

@CJ42 this needs to be better explained.

frozeman (changes requested):
We need to improve the explanations

LSPs/LSP-8-IdentifiableDigitalAsset.md
samuel-videau: What you mean by "not the token itself" ? This key is the replacement of `LSP8MetadataJSON:<address|uint256|bytes32>` right ? If yes then it's the key to store metadata for a specific token. So tokenURI make more sense for me and keep consistency with what existing in ERC721 (either baseURI & tokenURI)

samuel-videau: > `LSP8MetadataURI`, as off chain only meta data or related information is stored, and not the token itself. whats the thinking behind the `LSP8TokenURI` suggestion?

Ok but yeah I get your point, but let's stay consistent regarding the other key `LSP8MetadataBaseURI`, and call this one `LSP8MetadataTokenURI` or something like that

LSPs/LSP-8-IdentifiableDigitalAsset.md
samuel-videau: > What else than meta data can be stored under the URI?

Yup `LSP8MetadataTokenURI:<address|uint256|bytes32|string>` make more sense to keep consistent with the other `LSP8MetadataBaseURI`

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -68,142 +70,118 @@ The `tokenId` type can be one of the following possible enum values.
...
+This can be obtained as follow:
 
-#### LSP8TokenIdMetadata
+`keccak256('keccak256(utf8)')` = `0x`**`6f357c6a`**`956bf6b8a917ccf88cc1d3388ff8d646810d0393fe69ae7ee228004f`
 
-The description of the asset.
 
 ```json
 {
-    "name": "LSP4Metadata",
```
samuel-videau, line 183: Hmmm not sure we should remove `LSP4Metadata`, I mean this key will still be used to describe the contract metadata. Or maybe you consider we don't need to put it here as it's already in the `LSP4` doc? If it's the case then we at least need to reference it (same for LSP4TokenName actually (so could be good to reference to `LSP4` doc))

LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -68,142 +70,118 @@ The `tokenId` type can be one of the following possible enum values.
...
-#### LSP8MetadataAddress:TokenId
+When metadata JSON is created for a tokenId, the URI COULD be stored in the storage of the LSP8 contract.
 
-When a metadata contract is created for a tokenId, the address COULD be stored in the minting contract storage.
+The value stored under this data key is a tuple `(bytes4,string)` that contains the following elements:
 
-For construction of the Mapping data key see: [LSP2 ERC725Y JSON Schema > `keyType = Mapping`][LSP2#mapping]
+- `bytes4` = the 4 bytes identifier of the hash function used to generate the URI:
+    - if the tokenId is a hash (LSP8TokenIdType `3`): see details below.
+    - if the tokenId is any other LSP8TokenIdType: MUST be `0x00000000`.
```
samuel-videau, line 83: Well I still don't understand why we do that here, it was useful for the baseURI as we can t store JSONUrls. But here we could store a JSONUrl no matter the tokenIdType. Now it's a big regression as with this updated key, we loose verifiability on all the other tokenIds than the tokenId 3. Though it's adding consistency with the baseURI key

LSPs/LSP-8-IdentifiableDigitalAsset.md
CJ42: @samuel-videau it is mentioned further down below, in the section **LSP8 TokenId Metadata**

image

https://github.com/lukso-network/LIPs/blob/240d8baad573ef06cef8a5bdf6da5246f907967c/LSPs/LSP-8-IdentifiableDigitalAsset.md#lsp8-tokenid-metadata

LSPs/LSP-8-IdentifiableDigitalAsset.md
frozeman: because there is a big argument for changeable NFTs, where the meta data can be altered by an algorithm ,therefore producing different files. we shouldn't enforce hash referenced files, but rather give it as a feature.

samuel-videau: Alright I see

samuel-videau: Makes sense, as JSON url would force changeable NFTs to update the metadata keys all the time

frozeman:
Accidentally merged now, lets open a new PR for changes @CJ42

PR #206, opened 2023-04-26
[TRST-M2] Deviation from spec will result in dislocation of receiver delegate

CJ42:

In the specs of LSP0 + LSP9, add a warning note about the trimming rules for `LSP1UniversalReceiverDelegate:<bytes32>` type ID routing. To explain that the type id is trimmed on the right and the unique part of a data key SHOULD be on the left side of the `bytes32` (low order bytes).


YamenMerhi (approved):

CJ42:
I am marking this PR as draft as things need to be addressed.
We have to implement the behaviour to revert if no data is returned, and remove this from the specs.

PR #207, opened 2023-04-26
[TRST-M4] add notes about allowing the `universalReceiver(...)` function to return no data (void return)

CJ42:

Add a note in section in the LSP1 standard for the `universalReceiver(...)` function to specify that:

- it is possible for the function to return no data as `0x`
- if any `bytes` data is returned, it MUST be abi-encoded according to the ABI specs.


CJ42:
I am marking this PR as draft as things need to be addressed.
We have to implement the behaviour to revert if no data is returned, and remove this from the specs.

PR #208, opened 2023-04-26
[TRST-R3] Specification improvements for LSP0 and LSP20

CJ42:

## LSP0 + LSP9

- Improve the specs for LSP0 and LSP9 `universalReceiver(...)` function to describe that execution continues normally if the address stored under the data keys `LSP1UniversalReceiverDelegate` and `LSP1UniversalReceiverDelegate:<bytes32>` do not support the LSP1 interface ID.
- group the content for both LSP1 data keys in one big section paragraph.

## LSP20

- add _Requirements_ as subheaders for each methods.
- explicitly state that if the function being called does not return anything, the `callResult` value MUST be an empty `bytes`
Improve specs.
- add brief description under each method `lsp20VerifyCall(...)` and `lsp20VerifyCallResult(...)`

### DIFF
--- a/LSPs/LSP-0-ERC725Account.md
+++ b/LSPs/LSP-0-ERC725Account.md
@@ -387,13 +387,17 @@ This function is part of the [LSP1] specification, with additional requirements 
 
 - MUST emit a [`ValueReceived`] event before external calls if the function receives native tokens.
 
+- MUST emit a [UniversalReceiver] event if the function was successful.
+
 - If an `address` is stored under the data key attached below and this address is a contract:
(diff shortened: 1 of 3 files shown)


YamenMerhi:
I think now the explanation of universalReceiver lack more information, as it doesn't point that there is two universalReceiverDelegates, also mention attached below and they are far away from the description. I suggest to keep the old description and add the needed pts
Screen Shot 2023-05-02 at 9 59 09 AM

YamenMerhi (changes requested):
.

PR #211, opened 2023-06-06
Update LSP16 Specification

YamenMerhi:
- Finalize LSP16 Specification

### DIFF
--- a/LSPs/LSP-16-UniversalFactory.md
+++ b/LSPs/LSP-16-UniversalFactory.md
@@ -12,174 +12,588 @@ requires: EIP104, EIP155, EIP1167
 
 ## Simple Summary
 
-This standard defines a universal factory smart contract, that will allow to deploy different types of smart contracts using [CREATE2] opcode after being deployed with [Nick's method] in order to produce the same address on different chains.
+This standard defines a universal factory smart contract, that will allow to deploy different types of smart contracts using [CREATE2] opcode after being deployed with [Nick Factory] in order to produce the same address on different chains.
  
 ## Abstract
 
 This standard defines several functions to be used to deploy different types of contracts using [CREATE2] such as normal and initializable contracts. These functions take into consideration the need to initialize in the same transaction for initializable contracts. The initialize data is included in the salt to avoid squatting addresses on different chains.
 
-The same bytecode and salt will produce the same contract address on different chains, if and only if, the UniversalFactory contract was deployed on the same address on each chain, using [Nick's method] in our case.
+The same bytecode and salt will produce the same contract address on different chains, if and only if, the UniversalFactory contract was deployed on the same address on each chain, using [Nick Factory] in our case.
 
 ## Motivation
 
-Having a private key that controls an address makes it possible to control this address on different chains. This will allow people to check an address on a specific chain and send assets on the other chains assuming the same entity control the other addresses. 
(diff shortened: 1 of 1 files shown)


b00ste (approved):

LSPs/LSP-16-UniversalFactory.md
```diff
@@ -12,174 +12,705 @@ requires: EIP104, EIP155, EIP1167
...
+- `initializeCalldataMsgValue`: The value sent to the contract during initialization                                                                                               
+
+**Return:**
+
+- `contractCreated`: The address of the contract created.
+
+
+**Requirements:**
+
+- If value is associated with the contract creation, the constructor of the contract to deploy MUST be payable, otherwise the call will revert.
```
CJ42, line 544: suggested changes:

```suggestion
- If some value is transferred during the contract creation, the constructor of the contract to deploy MUST be payable, otherwise the call will revert.
```

LSPs/LSP-16-UniversalFactory.md
```diff
@@ -12,174 +12,705 @@ requires: EIP104, EIP155, EIP1167
...
+
+- If value is associated with the initialization call, the initialize function called on the contract to deploy MUST be payable, otherwise the call will revert.
```
CJ42, line 546: suggested changes:

```suggestion
- If some value is transferred during the initialization call, the initialize function called on the contract to deploy MUST be payable, otherwise the call will revert.
```

LSPs/LSP-16-UniversalFactory.md
```diff
@@ -12,174 +12,705 @@ requires: EIP104, EIP155, EIP1167
...
-function deployCreate2Proxy(address baseContract, bytes32 providedSalt, bytes calldata initializeCalldata) external payable returns (address);
+function deployERC1167Proxy(address implementationContract, bytes32 providedSalt) public returns (address);
 ```
 
-Deploys a [minimal proxy] using [CREATE2] with passed `baseContract` and a generated the salt. 
-If the initializeCalldata length is not zero, perform an external call on the contract created with `initializeCalldata` as payload and msg.value as value. initializeCalldata length is zero do not perform an external contract and the msg.value should be zero.
+Deploys an ERC1167 minimal proxy contract using the CREATE2 opcode.
 
-If the contract is initializable, the salt generated is the keccak256 of an initializable boolean, the initializeCallData, and the provided salt packed encoded. The initializable boolean is true in this case.
+The address where the contract will be deployed can be known in advance via the {computeERC1167Address} function.
```
CJ42, line 560: I suggest for these functions to use anchor links via `[computeERC1167Address](#computeERC1167Address)` instead of `{computeERC1167Address}`

LSPs/LSP-16-UniversalFactory.md
```diff
@@ -12,174 +12,705 @@ requires: EIP104, EIP155, EIP1167
...
 
-> `keccak256(abi.encodePacked(initializable, initializeCallData, providedSalt))`
+The `providedSalt` parameter is not used directly as the salt by the CREATE2 opcode. Instead, it is hashed via keccak256 with prepending a false boolean. See {generateSalt} function for more details.
```
CJ42, line 562: same here

CJ42 (approved):

PR #224, opened 2023-08-10
create LSP25 Execute Relay Call standard, extracting functions from LSP6 standard

### LINKED ISSUE #221: [LSP25] Execute Relay Call as its own standard (extracted from LSP6)
# Overview

Currently, the 3 x functions `getNonce`, `executeRelayCall` and `executeRelayCallBatch` are part of the LSP6 interface and embedded in the standard.

However, meta transactions in web3 are used in many different context, such as:

- voting (ask a 3rd party to cast your vote on your behalf, paying the gas for the transaction). See Uniswap example, or proxy voting in real life: [https://en.m.wikipedia.org/wiki/Proxy_voting#:~:text=Proxy voting is a form,enable a vote in absence](https://en.m.wikipedia.org/wiki/Proxy_voting#:~:text=Proxy%20voting%20is%20a%20form,enable%20a%20vote%20in%20absence).
- NFT Marketplaces that want to be built to accept meta transactions natively, without having to put a Key Manager in front.
- People wanting to build decentralised relayers
- Free NFTs with free mintings, given as gifts (e.g: Stickmen Toys: [https://stickmentoys.com/](https://stickmentoys.com/), or Warners Records UK that gave free musical NFTs). If you want to give NFTs for free, you can make the NFT to not have a price (free), but the transaction will always incurs a gas cost. Meta transaction can enable that.
- Any services that require a lot of transactions (streaming, microtransactions, gaming, defi, etc…)

By putting Meta Transaction in its own standard as a mew LSP, this offer a new separate small building block / component that devs can use to embed meta transactions in their own protocol / applications (instead to have to also opt for permissions management, which is not necessarely related and that they might not want).

# Questions

## Should LSP25 be a MUST in LSP6?

The question would be if:

- LSP25 is also part of LSP6 ➕ (as a COULD, like LSP20 is)
- LSP25 is not part of LSP6 ➖, and our LSP6KeyManager is simply an implementation contract built in to include also LSP20 + LSP25 so that it can be specifically used in front of a Universal Profile.

So far, while discussing on Monday with @frozeman, we mentioned that the specs of LSP6 have to specify that **it MUST includes the LSP25 standard + properly described with a short list of functions.**

## ⚠️ BREAKING CHANGE --> What would be the new version for signing an execute relay call?

So far for executeRelayCall, executors submit transactions signed. The data to sign for a valid `executeRelayCall` has to be formatted according to the following format (currently defined in LSP6 specs).

image

Since we are now moving this to its own standard, I am wondering what **should be the new 4th parameter** `LSP6_VERSION = 6`.

_option 1)_ Shouldn't this be changed to `LSP25_VERSION = 25` as a result?
_option 2)_ Or should this be kept as it is? The version is specific to the contract implementing the standard, like in this case LSP6?

On my side, I don't see a strong reason for option 2), but instead would favour option 1 to have the same approach as EIP191 but for different versions.
This way, contracts could keep the same LSP25 interface, but create different LSP standards for different signed data schemes (see screenshot below for EIP191)

**--> Important to be discussed**

https://eips.ethereum.org/EIPS/eip-191

image

# Technical references

These would be the changes in the interface of LSP6 + the standard interface of LSP25. (we take out these 3 functions out of LSP6 essentially).

## Interface of LSP6

```tsx
target()
execute()
executeBatch()

+ PermissionsVerified event
```

## Interface of LSP25

```tsx
getNonce()
executeRelayCall()
executeRelayCallBatch()
```

CJ42:

## ⚠️ BREAKING CHANGES

- Change interface Id of LSP6 from `0x38bb3cdb` to `0x627ca5d3`

## To be Discussed

- [x] Should we mentioned the LSP25 functions in the LSP6 specs, and add the extra requirements (recovered signer address MUST have the right permission for the `payload/callData`, emitted event `PermissionsVerified`.

Closes #221

### DIFF
--- a/LSPs/LSP-25-ExecuteRelayCall.md
+++ b/LSPs/LSP-25-ExecuteRelayCall.md
@@ -1,0 +1,276 @@
+---
+lip: 25
+title: Execute Relay Call
+author: Jean Cavallera (@CJ42)
+discussions-to: <URL>
+status: Draft
+type: LSP
+created: 2023-08-09
+requires: ERC165
+---
+
+**Table of Content**
+
+- [Simple Summary](#simple-summary)
+- [Abstract](#abstract)
+- [Motivation](#motivation)
+- [Specification](#specification)
+  - [Methods](#methods)
+    - [getNonce](#getnonce)
+    - [executeRelayCall](#executerelaycall)
+    - [executeRelayCallBatch](#executerelaycallbatch)
+  - [Signature Format](#signature-format)
+  - [Multi-Channel Nonces](#multi-channel-nonces)
+    - [What are multi-channel nonces?](#what-are-multi-channel-nonces-)
+    - [Problem of Sequential Nonces](#problem-of-sequential-nonces)
+    - [Benefits of multi-channel nonces](#benefits-of-multi-channel-nonces)
+    - [How nonces are represented across channels?](#how-nonces-are-represented-across-channels-)
+- [Implementation](#implementation)
+- [Interface Cheat Sheet](#interface-cheat-sheet)
+- [Copyright](#copyright)
+
+# Simple Summary
+
+This standard describes an interface and a signature scheme that can be used to integrate meta-transactions (transactions where users do not pay for the gas) inside a smart contract.
+
+# Abstract
+
+Meta transactions in web3 are used in many different context. From transferring buying, selling and transferring tokens and NFTs to voting systems, where delegates can submit transactions on behalf of other users and pay for the gas cost.
+
+# Motivation
+
+By having a common smart contract interface, applications and protocols can implement the LSP25 Execute Relay Call standard to start implementing gas-less transaction. This can be beneficial for protocols and applications like Digital Marketplaces or Universal Profile that want to onboard new users without requiring users to hold native tokens.
+
+# Specification
+
+**LSP25-ExecuteRelayCall** interface id according to [ERC165]: `0x5ac79908`
+
+Smart contracts implementing the LSP6 standard MUST implement the [ERC165] `supportsInterface(..)` function and MUST support the [ERC165] and LSP25 interface ids.
+
+## Methods
+
+### getNonce
+
+```solidity
+function getNonce(address signer, uint128 channel) external view returns (uint256)
+```
+
+Returns the latest nonce for a signer on a specific channel. A signer can choose a channel number arbitrarily and use this nonce to sign a payload that can be executed as a meta-transaction by any address via [executeRelayCall](#executeRelayCall) function.
+
+_Parameters:_
+
+- `signer`: the address of the signer of the transaction.
+- `channel` : the channel which the signer wants to use for executing the transaction.
+
+_Returns:_ `uint256` , the current nonce.
+
+Payloads signed with incremental nonces on the same channel for the same signer are executed in order. e.g, in channel X, the payload nb two signed with the second nonce will not be successfully executed until the payload nb one signed with the first nonce has been executed.
+
+Payloads signed with nonces on different channels are executed independently from each other, regardless of when they got executed or if they got executed successfully or not. e.g, the payload signed with the fourth nonce on channel X can be successfully executed even if the payload signed with the first nonce of channel Y:
+
+- was executed before.
+- was executed and reverted.
+
+> X and Y can be any arbitrary number between 0 and 2^128.
+
+Read [what are multi-channel nonces](#what-are-multi-channel-nonces).
+
+### executeRelayCall
+
+```solidity
+function executeRelayCall(bytes memory signature, uint256 nonce, uint256 validityTimestamps, bytes memory payload) external payable returns (bytes memory)
+```
+
+Allows anybody to execute a `payload` given they have a valid signature from a signer. The signature MUST be formed according to the [LSP25 Signature specification format](#signature-format).
+
+_Parameters:_
+
+- `signature`: bytes65 ethereum signature.
+- `nonce`: MUST be the nonce of the address that signed the message. This can be obtained via the `getNonce(address address, uint256 channel)` function.
+- `validityTimestamps`: Two `uint128` timestamps concatenated together. The first timestamp determines from when the payload can be executed, the second timestamp determines a deadlines after which the payload is not valid anymore. If validityTimestamps is `0`, the payload is valid at indefinitely at any point in time and the checks for the timestamps are skipped.
+- `payload`: The abi-encoded function call to be executed on the linked target contract.
+
+_Returns:_ `bytes` , the returned data as abi-decoded bytes of the call on ERC725 smart contract, if the call succeeded, otherwise revert with a reason-string.
+
+_Requirements:_
+
+- The address recovered from the signature and the digest signed MUST have **permission(s)** for the action(s) being executed. Check [Permissions](#permissions) to know more.
+- The nonce passed to the function MUST be a valid nonce according to the [multi-channel nonce](#what-are-multi-channel-nonces) section.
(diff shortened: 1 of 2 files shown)


LSPs/LSP-6-KeyManager.md
```diff
@@ -37,11 +36,16 @@ Storing the permissions at the core [ERC725Account] itself, allows it to survive
 
 ## Specification
 
-**LSP6-KeyManager** interface id according to [ERC165]: `0x38bb3cdb`.
+**LSP6-KeyManager** interface id according to [ERC165]: `0x627ca5d3`.
```
b00ste, line 39: interface id changed ninth latest `lip-smart-contracts` release to `0x66918867`

b00ste (approved):

LSPs/LSP-6-KeyManager.md
```diff
@@ -100,30 +104,45 @@ This function is part of the [LSP20-CallVerification] specification, with additi
...
+) 
+    external 
+    payable 
+    returns (bytes memory)
 ```
 
-Returns the latest nonce for a signer on a specific channel. A signer can choose a channel number arbitrarily and use this nonce to sign a payload that can be executed as a meta-transaction by any address via [executeRelayCall](#executeRelayCall)  function.
-
-_Parameters:_
+This function is part of the [LSP25-ExecuteRelayCall] specification, with additional requirements as follow.
```
JeneaVranceanu, line 121:
```suggestion
This function is part of the [LSP25-ExecuteRelayCall] specification, with additional requirements as follows.
```

LSPs/LSP-6-KeyManager.md
```diff
@@ -100,30 +104,45 @@ This function is part of the [LSP20-CallVerification] specification, with additi
...
+    uint256[] memory values, 
+    bytes[] memory payloads
+) 
+    external 
+    payable 
+    returns (bytes[] memory)
+```
 
-> X and Y can be any arbitrary number between 0 and 2^128.
+This function is part of the [LSP25-ExecuteRelayCall] specification, with additional requirements as follow.
```
JeneaVranceanu, line 142:
```suggestion
This function is part of the [LSP25-ExecuteRelayCall] specification, with additional requirements as follows.
```

LSPs/LSP-25-ExecuteRelayCall.md
```diff
@@ -0,0 +1,301 @@
...
+
+# Motivation
+
+By having a common smart contract interface, applications and protocols can implement the LSP25 Execute Relay Call standard to start implementing gas-less transaction. This can be beneficial for protocols and applications like Digital Marketplaces or Universal Profile that want to onboard new users without requiring users to hold native tokens.
+
+# Specification
+
+**LSP25-ExecuteRelayCall** interface id according to [ERC165]: `0x5ac79908`
+
+Smart contracts implementing the LSP6 standard MUST implement the [ERC165] `supportsInterface(..)` function and MUST support the [ERC165] and LSP25 interface ids.
```
JeneaVranceanu, line 48: Since we are linking ERC165 would it make sense to turn LSP6 into links as well?

JeneaVranceanu: It's only about convenience.

LSPs/LSP-25-ExecuteRelayCall.md
```diff
@@ -0,0 +1,301 @@
...
+Returns the latest nonce for a signer on a specific channel. A signer can choose a channel number arbitrarily and use this nonce to sign a payload that can be executed as a meta-transaction by any address via [executeRelayCall](#executeRelayCall) function.
+
+_Parameters:_
+
+- `signer`: the address of the signer of the transaction.
+- `channel` : the channel which the signer wants to use for executing the transaction.
+
+_Returns:_ `uint256` , the current nonce.
+
+Payloads signed with incremental nonces on the same channel for the same signer are executed in order. e.g, in channel X, the payload nb two signed with the second nonce will not be successfully executed until the payload nb one signed with the first nonce has been executed.
```
JeneaVranceanu, line 67: There's a typo?
`.. payload nb two ...`

Didn't get what `nb` means.

JeneaVranceanu: I see `nb` repeating further as well.

LSPs/LSP-25-ExecuteRelayCall.md
```diff
@@ -0,0 +1,301 @@
...
+    returns (bytes memory)
+```
+
+Allows anybody to execute a `payload` given they have a valid signature from a signer. The signature MUST be formed according to the [LSP25 Signature specification format](#signature-format).
+
+_Parameters:_
+
+- `signature`: A 65 bytes long ethereum signature.
+- `nonce`: MUST be the nonce of the address that signed the message. This can be obtained via the `getNonce(address address, uint256 channel)` function.
+- `validityTimestamps`: Two `uint128` timestamps concatenated together. The first timestamp determines from when the payload can be executed, the second timestamp determines a deadlines after which the payload is not valid anymore. If validityTimestamps is `0`, the payload is valid at indefinitely at any point in time and the checks for the timestamps are skipped.
```
JeneaVranceanu, line 98:
```suggestion
- `validityTimestamps`: Two `uint128` timestamps concatenated together. The first timestamp determines from when the payload can be executed, and the second timestamp determines a deadline after which the payload is no longer valid. If `validityTimestamps` is `0`, the payload is valid indefinitely at any point in time and the checks for the timestamps are skipped.
```

LSPs/LSP-25-ExecuteRelayCall.md
```diff
@@ -0,0 +1,301 @@
...
+- `payload`: The abi-encoded function call to be executed on the linked target contract.
```
JeneaVranceanu, line 99: LSP6 has specifications about the linked target contract while LSP25 says nothing about it. 
Abstract of LSP6 mentions linked target contract and LSP6 standard also documents `function target() external view returns (address)`.
Nothing here in LSP25 and again it was somewhat confusing.

JeneaVranceanu: It looks like you are mentioning here parts of LSP6 while LSP25 only `requires: ERC165`.
Since LSP6 depends on LSP25 it would make sense to reference bits and pieces of LSP25 within LSP6 but not the other way around.

It feels like `function target()` should be moved to this LSP as well.
What do you think?

LSPs/LSP-25-ExecuteRelayCall.md
```diff
@@ -0,0 +1,301 @@
...
+
+_Returns:_ `bytes` , the returned data as abi-decoded bytes of the call on ERC725 smart contract, if the call succeeded, otherwise revert with a reason-string.
```
JeneaVranceanu, line 101: Nothing up until this point suggested that the target contract should actually be an "ERC725 smart contract".
I'm highlighting this as this standard is generic and can be cross-chain.

LSPs/LSP-25-ExecuteRelayCall.md
CJ42: Will reword this

LSPs/LSP-25-ExecuteRelayCall.md
CJ42: It's an error, it should be LSP25, not LSP6. Thanks for flagging!

LSPs/LSP-25-ExecuteRelayCall.md
CJ42: @Jenea There should not be `target()` in this specs. I have re-worded the paragraph. Thanks for pointing it out :pray:

PR #237, opened 2023-09-22
Fix typos

omahs:
Fix typos


LSPs/LSP-16-UniversalFactory.md
```diff
@@ -718,7 +718,7 @@ An implementation can be found in the [lukso-network/lsp-smart-contracts](https:
 
 ## Security Consideration
 
-Knowing that deploying a contract using the UniversalFactory will allow to deploy the same contract on other chains with the same address, people should be aware and watch out to use contracts that doesn't have a logic that protect against replay-attacks.
+Knowing that deploying a contract using the UniversalFactory will allow to deploy the same contract on other chains with the same address, people should be aware and watch out to use contracts that dont't have a logic that protects against replay-attacks.
```
CJ42, line 721: It should be this actually

```suggestion
Knowing that deploying a contract using the UniversalFactory will allow to deploy the same contract on other chains with the same address, people should be aware and watch out to use contracts that don't have a logic that protects against replay-attacks.
```

CJ42 (approved):
Thanks! 👌🏻

PR #238, opened 2023-09-25
docs: update interface id for LSP7 and LSP8

magalimorin18:


CJ42 (changes requested):
Perfect many thanks for flagging this! 🙏 
Only thing we need is to make sure the interface cheat sheet at the bottom contains the same functions (names + params) than the one in:

- `ILSP7DigitalAsset.sol`: https://github.com/lukso-network/lsp-smart-contracts/blob/develop/contracts/LSP7DigitalAsset/ILSP7DigitalAsset.sol

- `ILSP8IdentifiableDigitalAsset.sol`: https://github.com/lukso-network/lsp-smart-contracts/blob/develop/contracts/LSP8IdentifiableDigitalAsset/ILSP8IdentifiableDigitalAsset.sol

CJ42:
In the LSP7 interface cheatsheet, the function `getOperatorsOf` is missing for instance.
There might be more

CJ42 (approved):

PR #249, opened 2023-10-18
LSP3/4: change hash to verification data

frozeman:
To allow for more flexible ways to very images and other sources referenced from smart contracts, we should change the property names in LSP3 and LSP4 for images and asset content to `verificationFunction`, `verificationData `

```js
{ // example of a verificationData based image verification
      "width": Number,
      "height": Number,
      "verificationFunction": 'keccak256(bytes)',
      "verificationData": 'string', // bytes32 hash of the image
      "url": 'string'
  },
  { // example of a signature based image verification
      "width": Number,
      "height": Number,
      "verificationFunction": 'ecdsa',
      "verificationData": 'string', // signer that signed the bytes of the image
      "verificationSource": 'string' // e.g url returning the signature of the signed image
      "url": 'string'
  },
```

This way it can also be used to verify server generated NFT images for example, as long as they have a `verificationSource` URL of some kind. One example is signature based verification of content, where the `verificationSource` returns the signed image datas, and the `verificationData` the signer address (could also be a UP address)

I know this is a bit of a change for projects like universal.page.
But this would enable server rendered and verifiable NFTs and other use cases.

Something we won't be able to add once all these tools are live and NFTs got deployed on mainnet

### DIFF
--- a/LSPs/LSP-3-Profile-Metadata.md
+++ b/LSPs/LSP-3-Profile-Metadata.md
@@ -76,49 +76,66 @@ The linked JSON file SHOULD have the following format:
         ],
         "tags": [ "string", "string", ... ], // tags related to the profile
         "avatar": [ // a 3D file avatar mostly in FBX and/or OBJ format, multiple file formats of an avatar can be added
-            {
-                "hashFunction": 'keccak256(bytes)',
-                "hash": 'string',
+            { // example of a hash based avatar verification
+                "verificationFunction": 'keccak256(bytes)',
+                "verificationData": 'string',
                 "url": 'string',
                 "fileType": 'string'
             },
-            ...
-            // OR link a DigitalAsset (NFT)
-            {  
+            { // example of a signature based avatar verification
+                "verificationFunction": 'ecdsa',
+                "verificationData": 'string',  // signer that signed the bytes of the image
+                "verificationSource": 'string' // e.g url returning the signature of the signed image
+                "url": 'string',
+                "fileType": 'string'
+            },
+            { // example of a NFT/smart contract based avatar
                 "address": Address, // the address of an LSP7 or LSP8
                 "tokenId": 32bytes  // (optional) if token contract is an LSP7
             }
         ]
         // below each image type SHOULD have different size of the same image, so that interfaces can choose which one to load for better loading performance
         "profileImage": [ // One image in different sizes, representing the profile.
-            {  
+            { // example of a hash based image verification
                 "width": Number,
                 "height": Number,
-                "hashFunction": 'keccak256(bytes)',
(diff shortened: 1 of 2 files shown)


frozeman:
CC @lykhonis

frozeman:
@Hugoo where we would need todo changes in order for this to be implemented?

lykhonis:
I am not as clear on signature verification of content part. A profile contract must approve an addition of a value (JSONURL) that adds a json stating some image is referenced inside a metadata. This action is effectively approving (signing) the metadata and the image. Image can be stored anywhere else, and it is approved/signed inherently by the profile because the metadata retrieved from the profile has the hash of the content (keccak256).

If this is not a case, could you expand on what do you mean by server side image verification?

Hugoo:
@frozeman 

- erc725.js -> https://github.com/ERC725Alliance/erc725.js/blob/develop/src/lib/getDataFromExternalSources.ts#L99

Any frontend/app that writes data to UP: Extension:
- https://github.com/lukso-network/universalprofile-extension/blob/95f3a1be57bfc96463a55b82bcb2216694e4253c/src/background/utils/profileIpfsUpload.ts#L81
- https://github.com/lukso-network/universalprofile-extension/blob/95f3a1be57bfc96463a55b82bcb2216694e4253c/src/shared/contracts/UniversalProfileContract.ts#L45
- Test dApp default profile value: https://github.com/lukso-network/universalprofile-test-dapp/blob/main/src/components/endpoints/SetData.vue#L13

+ community projects

frozeman:
@lykhonis 

This is allow server or AI generated image art, where the image constantly changes without additional transactions. E.g. there could be NFTs that are changing on every request.

This method allows for a method todo that in a verifiable and trustable way.
E.g. in this case: Only when the image is signed by the artists key, or any key referenced in `verificationData`, then its valid.

This also opens up the possibility of totally new verification methods, like zk proof based etc.

It makes the name generic and therefore possible for all kinds of verification types.

And that's not just for NFTs, but for anything.
Also unverified is ok with `verificationFunction = null`

lykhonis:
@frozeman signing an image content would imply ECDSA(KECCAK256(image content)). If image is changed, new signature thus new hash is required. Wouldn't this still require a key/value update due metadata keys being changed? This would still require to generate a hash of the image before signing it.

I am sure I am missing something, just not seeing this through yet.

frozeman:
Rather than adding new types of properties for every new use case, we better make these words generic, so they fit for many use cases. Now anything can be a verification function and data. Which will make this vastly more flexible.

As we are just about to launch mainnet, i think it's a change worth doing. before everyone actually starts to issue real NFTs

frozeman:
It should not be hard to make the changes for this before you guys launch, as will we.

PR #256, opened 2023-11-06
LSP3,4: changed to verification object

frozeman:
After discussing the properties, we found that it is better to put the verification into a new object, so its more upgradable in the future.
e.g.

```js
load for better loading performance
        "profileImage": [ // One image in different sizes, representing the profile.
            { // example of a hash based image verification
                "width": Number,
                "height": Number,
                "url": 'string',
                "verification": {
                    "method": 'keccak256(bytes)',
                    "data": 'string', // bytes32 hash of the image
                }
            },
            { // example of a signature based image verification
                "width": Number,
                "height": Number,
                "url": 'string',
                "verification": {
                    "method": 'ecdsa',
                    "data": 'string', // signer that signed the bytes of the image
                    "source": 'string' // e.g url returning the signature of the signed image
                }
            },
            { // example of a NFT/smart contract based image
                "address": Address, // the address of an LSP7 or LSP8
                "tokenId": 32bytes  // (optional) if token contract is an LSP7
            }
        ],
```

### DIFF
--- a/LSPs/LSP-3-Profile-Metadata.md
+++ b/LSPs/LSP-3-Profile-Metadata.md
@@ -77,15 +77,19 @@ The linked JSON file SHOULD have the following format:
         "tags": [ "string", "string", ... ], // tags related to the profile
         "avatar": [ // a 3D file avatar mostly in FBX and/or OBJ format, multiple file formats of an avatar can be added
             { // example of a hash based avatar verification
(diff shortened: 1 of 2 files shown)


frozeman:
@lykhonis This will be the necessary change before UP will go for mainnet, please make sure your app also is able to decode these new properties, if you use them.

PR #268, opened 2023-12-11
Update in LSP8 wording for schema and type to format

CJ42:


JeneaVranceanu (approved):
Approved but with the following changes to be applied: https://github.com/lukso-network/LIPs/pull/273

PR #269, opened 2023-12-13
chore: typos in LSP-2-ERC725YJSONSchema.md

JeneaVranceanu:
We had VeririableURI, VerifiableURL and VerifiedURL used interchangeably. Decided to update it to make it consistent.

As I see the type was initially planned to be named `VerifiableURI` and I highlight that it's not a `...URL`. That was the reason for changing entries from `...URL` to `...URI`.


CJ42:
Let's double check with @frozeman as the old ones were **ASSETURL** and **JSONURL**

image

image

YamenMerhi (approved):

Hugoo (approved):

PR #271, opened 2023-12-14
fix: changing JSONURL uses to VerifiableURI in LSP-3-Profile-Metadata.md

JeneaVranceanu:
Updated LSP3 to use VerifiableURI instead of JSONURL.

It doesn't seem that we need to change anything else since those schemas have `keyType` Singleton. `key` value stays the same.


magalimorin18 (approved):
nice
Also out of the scope of this PR but maybe could be tackled within the same Pr because it s same repo, there is a lot of spelling mistakes in the verifiableURI docs. If not addressed here, i can open a new PR for it

https://github.com/lukso-network/LIPs/blob/main/LSPs/LSP-2-ERC725YJSONSchema.md#verifiableuri

Hugoo (approved):

JeneaVranceanu:
@magalimorin18 Good catch! As this PR is already merged I'll open a new one with spelling mistakes fixed. Will tag you in discord 🙂

PR #274, opened 2024-01-02
[I-04] Inconsistencies in the LSP-7 specification

YamenMerhi:
#### authorizedAmountFor
- *Parameters* 
   - Add `tokenOwner` missing parameter
   
#### increaseAllowance

- *Requirements* 
  - `operator` cannot be calling address.
  - `operator` cannot be the zero address.

#### decreaseAllowance

- *Requirements* 
  - `operator` cannot be calling address.
  - `operator` cannot be the zero address.
  - `subtractedAmount` must be less than the `operator`'s current allowance.

#### transferBatch

- *Requirements* 
  - `from`, `to`, `amount`, `force`, and `data` lists are the same length.

#### Interface Cheat Sheet

- `transferBatch()` - `force` parameter `bool` array --> `bool[] memory force`.
- `batchCalls()` - `data` parameter `calldata`. --> `memory`. 

<hr>

#### transfer

- *Requirements*
  - `from` and `to` cannot be the same address.

#### transferBatch

- *Requirements* 
  - `from` and `to` cannot be the same address.

### DIFF
--- a/LSPs/LSP-7-DigitalAsset.md
+++ b/LSPs/LSP-7-DigitalAsset.md
@@ -175,6 +175,8 @@ _Parameters:_
 _Requirements:_
 
 - `operator`'s original allowance cannot be zero.
+- `operator` cannot be the zero address.
+- `operator` cannot be the calling address.
 
 **LSP1 Hooks:**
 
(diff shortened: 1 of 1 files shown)


LSPs/LSP-7-DigitalAsset.md
```diff
@@ -175,6 +175,8 @@ _Parameters:_
 _Requirements:_
 
 - `operator`'s original allowance cannot be zero.
+- `operator` cannot be the zero address.
+- `operator` cannot be the calling address.
```
CJ42, line 179: Maybe this wording makes more sense?

```suggestion
- `operator` cannot be the caller's address.
```

LSPs/LSP-7-DigitalAsset.md
```diff
@@ -220,6 +228,7 @@ Operators can send and burn tokens on behalf of their owners. The tokenOwner is
 _Parameters:_
 
 - `operator` the address to query operator status for.
+- `tokenOwner` the address whose token allowance is being queried.
```
CJ42, line 231: This looks incorrect to me.
The word allowance is usually used when talking about the operator. I would suggest the following re-wording for these two points:

```
- `operator` the address to query the operator's allowance for.
- `tokenOwner` the address whose operator's allowance is being queried for.
```

CJ42 (changes requested):

CJ42 (approved):

PR #284, opened 2024-01-22
fix: Add documentation for url-encoding requirements

richtera:


LSPs/LSP-8-IdentifiableDigitalAsset.md
```diff
@@ -518,22 +518,27 @@ _Requirements:_
...
 
 ⚠️ Similarly for the `bytes32` version of tokenId format (`3` and `4`), the tokenId part MUST be lowercase.
 
-⚠️ However, for the `string` version of the tokenId (`2`), the tokenId part MUST be in the same casing as defined by the string that represents the tokenId. Otherwise, changing the casing is equivalent to using a different tokenId.
+⚠️ However, for the `string` version of the tokenId (`2`), the tokenId part MUST be in the same casing as defined by the string that represents the tokenId. Otherwise, changing the casing is equivalent to using a different tokenId. The tokenId is not limited in casing but needs to contain valid UTF-8 and needs to be encoded using url-encoding i.e. encodeURI(). For example: A tokenId containing `something/Veryሴ⍅ Cool` should be encoded as `something/Very%E1%88%B4%E2%8D%85%20Cool` as in the following example.
```
CJ42, line 527:
```suggestion
⚠️ However, for the `string` version of the tokenId (`2`), the tokenId part MUST be in the same casing as defined by the string that represents the tokenId. Otherwise, changing the casing is equivalent to using a different tokenId. The tokenId is not limited in casing but **needs to contain valid UTF-8** and needs to be encoded using url-encoding i.e. encodeURI(). For example: A tokenId containing `something/Veryሴ⍅ Cool` should be encoded as `something/Very%E1%88%B4%E2%8D%85%20Cool` as in the following example.
```

CJ42: Suggest requirement a quote block as requirement like below to highlight better:

> **Requirements:**
> - The tokenId is not limited in casing but **needs to contain valid UTF-8**
> - The tokenId MUST be encoded using url-encoding (_e.g: `encodeURI()`_).

_Example:_

A tokenId containing `something/Veryሴ⍅ Cool` should be encoded as `something/Very%E1%88%B4%E2%8D%85%20Cool` as in the following example.

```js
> baseURL + encodeURI("something/Very\u1234\u2345 Cool")
'https://mybase.uri/something/Very%E1%88%B4%E2%8D%85%20Cool'
```

CJ42 (commented):
added review comments

CJ42 (approved):

PR #294, opened 2024-05-16
fix: missing ERC173  in 'requires' section of LSP8

JeneaVranceanu:


frozeman:
Isn't it using own able 2 step @CJ42

CJ42:
> Isn't it using own able 2 step @CJ42

No it is not, since it is inheriting ERC725Y

PR #295, opened 2024-05-30
fix: outdated links for lsp-smart-contracts replaced with new ones

JeneaVranceanu:
Noticed that old links to smart contract implementation were no longer working.


CJ42 (approved):
Nice thanks!

PR #304, opened 2024-09-14
draft: LIP-27 - Token Bound Profiles Proposal

ValentineCodes:
LIP-27 introduces a standard for creating Token Bound Profiles (TBP) using LSP8 Identifiable Digital Assets and Universal Profiles (LSP0) on the Lukso blockchain. This proposal allows each LSP8 token to be associated with a Universal Profile, enabling token-based ownership and interactions.


JordyDutch:
👏🏼🚀🤝🏼

tantodefi:
For consideration with 'Ownership Cycles'

We could potentially prevent the LSP8 being sent to the associated UP of the Token Bound Profile - by adding a check/override in the transfer function - although I'm not sure if this is desirable as we should try to keep the interfaces of the Profiles as close to the same as possible.

ValentineCodes:
> For consideration with 'Ownership Cycles'
> 
> We could potentially prevent the LSP8 being sent to the associated UP of the Token Bound Profile - by adding a check/override in the transfer function - although I'm not sure if this is desirable as we should try to keep the interfaces of the Profiles as close to the same as possible.

This seems like a step in the right direction, but it may not work well with existing NFTs.

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+status: Draft
+type: LSP
+category: Core
+created: 2024-09-14
+requires: LSP0, LSP8
+---
+
+## Simple Summary
+
+LIP-27 introduces a standard for creating Token Bound Profiles (TBP) using **LSP8 Identifiable Digital Assets** and **Universal Profiles (LSP0)** on the Lukso blockchain. This standard enables LSP8 tokens to be associated with a Universal Profile, facilitating token-based ownership and interaction.
```
YamenMerhi, line 14:
```suggestion
LIP-27 introduces a standard for creating Token Bound Profiles (TBP) using **LSP8 Identifiable Digital Assets** and **Universal Profiles (LSP0)**. This standard enables LSP8 tokens to be associated with a Universal Profile, facilitating token-based ownership and interaction.
```

YamenMerhi (commented):
Initial thoughts: 
- Achieve almost the same functionalities of ERC6551
- Bounding Account to tokenId is possible using this method, also with having accounts as direct bytes32 tokenId, but that is more for advanced usecases or a start of an NFT with account. The LSP-27 can be helpful for normal tokenId representing as numbers, hex, etc .. to be later associated with an account.
- Any Implementation ? The implementation needs to be reviewed carefully especially on the part of overriding ownership to be accessible by the token owner.
- Better to allow full profile creation with bytecode, and probably initialization calldata ?  
- Wording should be independent from LUKSO Blockchain, better to use "LSPs" to refer to the ecosystem of LSPs.
- More clarity about the registry: where is it deployed, how, etc ..

frozeman:
What is the purpose of the chain ID in the profile creation function?

ValentineCodes:
Ensures that the profile is uniquely tied to a specific network. However, as the LSPs are only for Lukso Blockchain, do you think this is unnecessary?

ValentineCodes:
> What is the purpose of the chain ID in the profile creation function?

Ensures that the profile is uniquely tied to a specific network. However, as the LSPs are only for Lukso Blockchain, do you think this is unnecessary?

frozeman:
> Ensures that the profile is uniquely tied to a specific network. However, as the LSPs are only for Lukso Blockchain, do you think this is unnecessary?

LSPs work on any chain, but in order that this LSP8 can set the owner of a profile, both the smart contract of the profile, as well as this LSP8 need to be on the same chain.
Hence why i am not sure what the chain ID is doing here.

Do i understand this right, that this is a simple LSP8 that deploys profiles, which then use the owner of the LSP8 token ID as owner of itself?

If thats the case, here are a few points:
- If the owner is changed by the transfer of the LSP8, then no keymanager can control this profile, meaning it looses all the special powers of LSP6 (permissions etc)
- The tokenId should be an `address`, which is the address of the profile

ValentineCodes:
> Initial thoughts:
> 
> * Achieve almost the same functionalities of ERC6551
> * Bounding Account to tokenId is possible using this method, also with having accounts as direct bytes32 tokenId, but that is more for advanced usecases or a start of an NFT with account. The LSP-27 can be helpful for normal tokenId representing as numbers, hex, etc .. to be later associated with an account.
> * Any Implementation ? The implementation needs to be reviewed carefully especially on the part of overriding ownership to be accessible by the token owner.
> * Better to allow full profile creation with bytecode, and probably initialization calldata ?
> * Wording should be independent from LUKSO Blockchain, better to use "LSPs" to refer to the ecosystem of LSPs.
> * More clarity about the registry: where is it deployed, how, etc ..

Thank you for reviewing this. I’ll update the documentation based on your suggestions.

ValentineCodes:
> > Ensures that the profile is uniquely tied to a specific network. However, as the LSPs are only for Lukso Blockchain, do you think this is unnecessary?
> 
> LSPs work on any chain, but in order that this LSP8 can set the owner of a profile, both the smart contract of the profile, as well as this LSP8 need to be on the same chain. Hence why i am not sure what the chain ID is doing here.
> 
> Do i understand this right, that this is a simple LSP8 that deploys profiles, which then use the owner of the LSP8 token ID as owner of itself?
> 
> If thats the case, here are a few points:
> 
> * If the owner is changed by the transfer of the LSP8, then no keymanager can control this profile, meaning it looses all the special powers of LSP6 (permissions etc)
> * The tokenId should be an `address`, which is the address of the profile

You make a good point. Using the tokenId as the address of the profile would be a great idea for newly created NFTs. However, how would this approach work for existing NFTs?

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+        address profile,
+        address indexed implementation,
+        bytes32 salt,
+        uint256 chainId,
+        address indexed tokenContract,
+        uint256 indexed tokenId
+    );
+
+    /**
+     * @dev The registry must revert with ProfileCreationFailed error if the CREATE2 operation fails.
```
CJ42, line 61: It should be precised in the standard if CREATE2 opcode MUST or COULD be used.

Meaning if it's mandatory to deploy UPs using CREATE2 or not

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * @dev The registry must emit the LSP27ProfileCreated event upon successful profile creation.
+     */
+    event LSP27ProfileCreated(
+        address profile,
+        address indexed implementation,
+        bytes32 salt,
+        uint256 chainId,
+        address indexed tokenContract,
+        uint256 indexed tokenId
```
CJ42, line 57: I think if you use LSP8TokenIdFormat = Address + CREATE2, the tokenId would be the address of the UP. So this tokenId parameter would not be necessary. The parameter `address profile` emitted in the event would act at the same time as the tokenId.

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+
+- **Purpose**: Initializes the Universal Profile, allowing for funding upon deployment and setting the LSP8 token as the profile's owner.
+
+#### Owner Function
+
+```solidity
+function owner() public view virtual returns (address);
+```
+
+- **Purpose**: The ownership of the Universal Profile is determined by calling the `tokenOwnerOf()` function from the LSP8 contract. Ownership is transferred along with the token transfer.
```
CJ42, line 114: > The ownership of the Universal Profile is determined by calling the `tokenOwnerOf()` function from the LSP8 contract.

This function expects a parameter. What would be the tokenId passed here when calling the `owner()` function? Where is the tokenId stored in the UP? I see it in the constructor but nothing else is mentioned about that.

> Ownership is transferred along with the token transfer.

Then this would mean if you want to have this exact behaviour, you would need to:
1. Override the _acceptOwnership function to call the LSP8 contract to transfer the tokenId to the new owner (challenging as here the caller would be the UP, so some custom logic would need to be implemented in your LSP8 version).

2. Override the `transfer(...)` function in LSP8 to also transfer ownership of the UP to the new owner. This would not work neither, the call might fail as the caller would not be the tokenId owner but the tokenContract. Therefore more custom logic would need to be implemented on the UP side for this.

CJ42, line 114: > The ownership of the Universal Profile is determined by calling the `tokenOwnerOf()` function from the LSP8 contract.

This function expects a parameter. What would be the tokenId passed here when calling the `owner()` function? Where is the tokenId stored in the UP? I see it in the constructor but nothing else is mentioned about that.

> Ownership is transferred along with the token transfer.

Then this would mean if you want to have this exact behaviour, you would need to:
1. Override the _acceptOwnership function to call the LSP8 contract to transfer the tokenId to the new owner (challenging as here the caller would be the UP, so some custom logic would need to be implemented in your LSP8 version).

2. Override the `transfer(...)` function in LSP8 to also transfer ownership of the UP to the new owner. This would not work neither, the call might fail as the caller would not be the tokenId owner but the tokenContract. Therefore more custom logic would need to be implemented on the UP side for this.

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### Transfer Ownership
+
+```solidity
+function transferOwnership(address newOwner) public;
+```
+
+- **Purpose**: Ownership of the Universal Profile is transferred by transferring the associated LSP8 token, instead of updating the state within the Universal Profile contract.
```
CJ42, line 122:
```suggestion
- **Purpose**: Ownership of the Universal Profile is transferred by transferring the associated LSP8 tokenId, instead of updating the state within the Universal Profile contract.
```

But you would still need to update the `_owner` in the state as well. How would that become different? Otherwise, this would override the standard behaviour (which is from LSP14 Ownable 2 Steps standard that LSP0 uses, so making it less compliant)

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+## Rationale
+
+The rationale for this proposal stems from the increasing complexity of digital assets and the need for a flexible, token-bound profile system that enables more nuanced token ownership and interaction capabilities.
+
+By leveraging **LSP8 Identifiable Digital Assets** and **LSP0 Universal Profiles**, we can create a framework for **Token Bound Profiles** that fits seamlessly within the Lukso ecosystem.
+
+This approach offers several advantages:
+
+- **Seamless Integration**: Token Bound Profiles build upon existing Lukso standards (LSP0 and LSP8), making them compatible with the ecosystem's existing infrastructure.
+- **Programmable Ownership**: By binding ownership to LSP8 tokens, profiles are automatically controlled by token holders without requiring manual intervention.
```
CJ42, line 141: There is a nuance here. By doing so, UP would not be able to be held by custom owner contracts (a Key Manager, a Multi sig, etc...).

This would be very restrictive, as as a result, Token Bound profile would not be able to benefit of gas less transactions.

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+
+#### Renounce Ownership
+
+```solidity
+function renounceOwnership() public;
+```
+
+- **Purpose**; This function permanently deletes the token contract and ID from storage, leaving the profile without an owner.
+
+## Rationale
```
CJ42, line 132: I do not find this section clear. This is an important one for proposing a new standard. 

It does highlight generically the potential benefits (although I find these more generic statements than specific advantages or problems being solved by the standard).

I would suggest to review this section carefully if looking to push this standard, as the rationale is what motivate the developers and projects in the ecosystem to adopt the standard, as it solves their specific needs.

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+
+This approach offers several advantages:
+
+- **Seamless Integration**: Token Bound Profiles build upon existing Lukso standards (LSP0 and LSP8), making them compatible with the ecosystem's existing infrastructure.
+- **Programmable Ownership**: By binding ownership to LSP8 tokens, profiles are automatically controlled by token holders without requiring manual intervention.
+- **Flexible Interactions**: This system allows tokens to interact with decentralized applications (dApps) and manage assets autonomously via their associated profiles.
+
+## Backwards Compatibility
+
+LIP-27 is fully compatible with existing **LSP0** and **LSP8** standards. It introduces no breaking changes to the Universal Profile or Identifiable Digital Asset contracts.
```
CJ42, line 146: This is a very strong statement. By the look of the function section above, it seems to break partially LSP14 which is part of LSP0. Although this could be debatable to the extent.

The bigger problem is that this standard proposal breaks composability over "who can be the owner of the UP". In the sense "control the UP". Ownership and control (via the `owner()` function) are two very different things

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+
+- **LSP0 (Universal Profile)**: The modifications to the Universal Profile contract under LIP-27 are backward compatible and do not alter the existing interface. The contract logic is adapted to incorporate LSP8 token ownership without disrupting the standard functionality.
```
CJ42, line 148: Yes that's correct from the public interface point of view. So this statement nuances my previous comment above.

However, it does disrupt the standard functionalities.

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+
+Preventing fraud is outside the scope of this proposal.
+
+### Ownership Cycles
+
+All assets held in a token bound account may become inaccessible if an ownership cycle is created. For example:
+
+- An LSP8 token could be transferred to its own token bound account. If this occurs, both the LSP8 token and all assets stored in the token bound account would be permanently inaccessible, as the token bound account is incapable of executing a transaction that transfers the LSP8 token.
+
+Application clients and account implementations wishing to adopt this proposal are encouraged to implement measures that limit the possibility of ownership cycles.
```
CJ42, line 182: This must be enforced at the smart contract level. But this my not be applicable as I believe LSP14 Ownable 2 Steps mention in the specs that it cannot transfer ownership to self (to be verified)

LIPs/lip-27.md
```diff
@@ -0,0 +1,196 @@
...
+
+Consider the following potential scam:
+
+- Alice owns an LSP8 token X, which owns token bound account Y.
+- Alice deposits 10 LYX into account Y.
+- Bob offers to purchase token X for 11 LYX via a decentralized marketplace, assuming he will receive the 10 ETH stored in account Y along with the token.
+- Alice withdraws 10 LYX from the token bound account and immediately accepts Bob’s offer.
+- Bob receives token X, but account Y is empty.
+
+To mitigate fraudulent behavior by malicious account owners, decentralized marketplaces should implement protection against these sorts of scams. Here are a few mitigation strategies to consider:
```
CJ42, line 167: Interesting section 👀

CJ42 (changes requested):
Thanks for submitting this LSP proposal @ValentineCodes 
 
I have added my review comments.

I am not agreeing on this standard  and I think it should not be merged and accepted for two big reasons:

1) this is a very custom implementation linking LSP8 to a UP (or LSP0 more generically). We should go away of proposing standards for custom implementations logic. The aim of a standard is to be generic purpose for a wide range of use cases. Here the custom logic around the ownership management make it very tightly coupled to LSP8 functionalities. Meaning its "defining a custom logic as something standard". The logic should define the MUST and MUST not for standard behaviour (to describe what the function MUST and MUST NOT do), but the standard shouldn't define HOW it does it (here it is the case, it defines "it should call the `tokenOwner(bytes32)` function to retrieve the `owner()`. Therefore it is too enforcing.

2) this standard breaks composability of a Universal Profile when it comes to ownership management. With this standard, a token bound profile:
- could not be owned by a Key Manager anymore. Therefore users of a token bound profile would not be able to set permissions in their UP + would not have access to gas less transactions anymore. **This would be a huge problem in this ecosystem** as user would then need to pay for gas for their own transaction and this goes against one of the main ideas around LUKSO and user friendly experience.

In the meantime for 2), no other type of owner contract would be able to own a UP (eg: a Multisig, a Governor contract, etc...). Or if not fully, would be limited.

My suggestion would be to use the existing LSP8 standard with LSP8TokenIdFormat as Address and override the transfer functionality internally as explained in my comment: deploying a UP on minting, and renouncing ownership of the UP on burn (burning is more challenging tho).

Finally this standard is very restrictive in the sense that the holder of the NFT (= token bound profile) is also the owner of the UP (= the one who controls the profile and can setData, execute, etc...). What about if an organisation want to create token bound profiles and be the owner of the tokenIds (= part of their assets portfolio), but want to allow their users to use these UP (by renting these UP to users via a subscription model) so their address are the `owner()`.

These are my reasons why I would not upvote this standard to get it merged into the repo and pushed forward to the ecosystem. However, you can always use this standard and your own implementation regardless if it has been merged, accepted or not (examples are @lykhonis and @jakeprins who proposed the LSP18 standard and used it already on universal.page even before the standard was finally accepted and merged).

CJ42:
> > > Ensures that the profile is uniquely tied to a specific network. However, as the LSPs are only for Lukso Blockchain, do you think this is unnecessary?
> 
> > 
> 
> > LSPs work on any chain, but in order that this LSP8 can set the owner of a profile, both the smart contract of the profile, as well as this LSP8 need to be on the same chain. Hence why i am not sure what the chain ID is doing here.
> 
> > 
> 
> > Do i understand this right, that this is a simple LSP8 that deploys profiles, which then use the owner of the LSP8 token ID as owner of itself?
> 
> > 
> 
> > If thats the case, here are a few points:
> 
> > 
> 
> > * If the owner is changed by the transfer of the LSP8, then no keymanager can control this profile, meaning it looses all the special powers of LSP6 (permissions etc)
> 
> > * The tokenId should be an `address`, which is the address of the profile
> 
> 
> 
> You make a good point. Using the tokenId as the address of the profile would be a great idea for newly created NFTs. However, how would this approach work for existing NFTs?

This is a good point. I think some basic examples in docs.lukso.tech of how LSP8 with Token Id Format Address could be used and the address being a UP (already deployed, or deployed at the time of minting) would be very valuable examples of use cases and code examples.

ValentineCodes:
> Thanks for submitting this LSP proposal @ValentineCodes 
>  
> I have added my review comments.
> 
> I am not agreeing on this standard  and I think it should not be merged and accepted for two big reasons:
> 
> 1) this is a very custom implementation linking LSP8 to a UP (or LSP0 more generically). We should go away of proposing standards for custom implementations logic. The aim of a standard is to be generic purpose for a wide range of use cases. Here the custom logic around the ownership management make it very tightly coupled to LSP8 functionalities. Meaning its "defining a custom logic as something standard". The logic should define the MUST and MUST not for standard behaviour (to describe what the function MUST and MUST NOT do), but the standard shouldn't define HOW it does it (here it is the case, it defines "it should call the `tokenOwner(bytes32)` function to retrieve the `owner()`. Therefore it is too enforcing.
> 
> 2) this standard breaks composability of a Universal Profile when it comes to ownership management. With this standard, a token bound profile:
> - could not be owned by a Key Manager anymore. Therefore users of a token bound profile would not be able to set permissions in their UP + would not have access to gas less transactions anymore. **This would be a huge problem in this ecosystem** as user would then need to pay for gas for their own transaction and this goes against one of the main ideas around LUKSO and user friendly experience.
> 
> In the meantime for 2), no other type of owner contract would be able to own a UP (eg: a Multisig, a Governor contract, etc...). Or if not fully, would be limited.
> 
> My suggestion would be to use the existing LSP8 standard with LSP8TokenIdFormat as Address and override the transfer functionality internally as explained in my comment: deploying a UP on minting, and renouncing ownership of the UP on burn (burning is more challenging tho).
> 
> Finally this standard is very restrictive in the sense that the holder of the NFT (= token bound profile) is also the owner of the UP (= the one who controls the profile and can setData, execute, etc...). What about if an organisation want to create token bound profiles and be the owner of the tokenIds (= part of their assets portfolio), but want to allow their users to use these UP (by renting these UP to users via a subscription model) so their address are the `owner()`.
> 
> These are my reasons why I would not upvote this standard to get it merged into the repo and pushed forward to the ecosystem. However, you can always use this standard and your own implementation regardless if it has been merged, accepted or not (examples are @lykhonis and @jakeprins who proposed the LSP18 standard and used it already on universal.page even before the standard was finally accepted and merged).

Thanks a lot for taking the time to respond. These are great inputs and you're right. Not every concept should be a standard. We'll review our approach with this in mind 🤝

frozeman:
Additionally, I want to mention here that I don't think a custom implementation should be part of a standard specification. What you want to define is the functions and their meaning, and not the code directly, and that seemed to me too specific on the implementation front. 

But I fully agree here with @CJ42 . If we make profiles owned like this, then it's not really a profile that anybody would ever want to own, because it's not really a cell phone profile. 

Tradeability of profiles could be easily built by just simply building a trading contract that then is authorized in each of the profiles that can be traded. It can call the transfer ownership functions to auto transfer ownership. That will be the proper composable way to do it.

PR #315, opened 2026-01-08
feat: add LSP29 Encrypted Assets standard

b00ste:

Introduces LSP29 for storing encrypted, token-gated digital assets on Universal Profiles.

## ERC725Y Data Keys
- `LSP29EncryptedAssets[]` - Array of VerifiableURIs
- `LSP29EncryptedAssetsMap` - O(1) lookup for latest/specific versions
- `LSP29EncryptedAssetRevisionCount` - Revision tracking per content ID

## Key Features
- Append-only array with full version history
- Single mapping supporting both "latest" and "specific version" lookups
- Creator-chosen content IDs that persist across revisions

### DIFF
--- a/LSPs/LSP-29-EncryptedAssets.md
+++ b/LSPs/LSP-29-EncryptedAssets.md
@@ -1,0 +1,382 @@
+---
+lip: 29
+title: Encrypted Assets
+author: b00ste
+discussions-to:
+status: Draft
+type: LSP
+created: 2025-01-08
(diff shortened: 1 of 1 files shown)


frozeman:
@b00ste  We should get back to this discussion, but I'm feeling that the data keys are not fully clear yet. The spec doesn't really show the data keys properly, and also the data key cheat sheet is missing. I think that needs to be more closely resembling LSP3 or LSP4 in the spec design. To be better readable

PR #318, opened 2026-03-01
feat: create LSP-31-MultiStorageURI.md

b00ste:


frozeman:
@b00ste How is this different to a verifiable UI, and also I guess that would make more sense to be added to LSP2 then as a new section?

